Issue #45, opened 2022-07-15
Switch to PNPM


IPWright83:
Remove `npm` and switch to `pnpm`

Issue #44, opened 2022-07-15
Fix GitHub Actions


IPWright83:
- [x] Fix all the pipelines
- [x] Protect the master branch
- [x] Sort out auto releases to NPM

Issue #41, opened 2022-07-07
Setup peer dependencies


IPWright83:
This project has been bootstrapped with `create-react-app`, but probably shouldn't. There are a bunch of dependencies that need moving to peer dependencies to ensure the library can be consumed correctly without version mis-matches.

PR #47, opened 2022-07-21
Update deps

IPWright83:
- Switches to `pnpm`
- Switched to using peer dependencies
- Fixed a broken story


IPWright83:
Resolves #45 
Resolves #44 
Resolves #41

PR #221, opened 2026-07-18
Add Pie, Donut and StackedDonut charts

IPWright83:
Introduces a polar charting stack that mirrors the XYChart architecture: a generic `<PieChart>` wrapper (Chart + TooltipOverlay + LegendOverlay) composed with `<Pie>`/`<Donut>`/`<StackedDonut>` plot components, following the same layering, HOC (`withPolarPlot` alongside `withSVG`/`withCanvas`), and SVG/Canvas dual-rendering conventions as the existing Bar/Column/Area plots.

Rebased on top of #219 (the removal of the `@chart-io/core` renderers abstraction), so the render/tooltip/focus logic for these plots lives directly in the React `Base` components and hooks, matching the current architecture:

- `@chart-io/core`: a generic `renderArc` canvas primitive (alongside `renderCircle`/`renderRect`, wired into `renderElements`) so Pie/Donut slices support Canvas + virtual-canvas hit-testing like other plots. Also adds `d3-hierarchy` (used client-side for the StackedDonut radial partition).
- `@chart-io/react`:
  - `<PieChart>` — the polar equivalent of `<XYChart>`
  - `<Pie>` / `<Donut>` — single-ring plots (`Donut` defaults `innerRadius` to `0.6`)
  - `<StackedDonut>` — builds a 2-level radial partition (`d3.hierarchy`/`d3.partition`) so the outer ring subdivides each inner-ring category proportionally
  - Reuses the existing Legend/Tooltip overlays and hooks (`useLegendItems`) rather than introducing new ones
- Storybook stories/docs under "PieCharts", and unit tests (SVG + Canvas) covering rendering, interactivity and event dispatch.
- Changeset included for the version bump.


packages/core/src/canvas/renderElements.ts
```diff
@@ -38,6 +39,10 @@ export function renderElements(
                 renderRect(context, node, overrideColor);
                 break;
 
+            case "PATH":
+                renderArc(context, node, overrideColor);
```
IPWright83: I'm not sure we can always assume a path is an arc. Maybe we need to make renderArc more generic and call it renderPath instead? We'd need to ensure it'll animate appropriately though

packages/react/src/lib/components/Plots/Pie/Donut/Donut.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Pie } from "../Pie";
+
+export type IDonutProps = IPieProps;
+
+/**
+ * Represents a Donut plot. This is a `<Pie>` with a hole in the middle, defaulting `innerRadius` to `0.6`
+ * @param  props       The set of React properties
+ * @return             The Donut plot component
+ */
+export function Donut({ innerRadius = 0.6, ...props }: IDonutProps) {
```
IPWright83: This should be inverted. Pie plots call a DonutPlot instead with 0 inner radius as it doesn't make sense giving an inner radius to a Pie segment.

packages/core/src/canvas/renderElements.ts
IPWright83: Good catch. Fixed by adding a `renderPath.ts` dispatcher in front of `renderArc`: it looks up the renderer via a `data-path-type` attribute stamped on the node, so a `<path>` only gets arc treatment because it's explicitly tagged as one — no more blanket assumption.

I did look at making it fully generic by parsing the `d` attribute directly (any path, not just arcs), but that turned out to be a real trap: d3-shape serializes a full-circle arc to SVG using deliberately near-but-not-exactly-coincident endpoints (to work around SVG's `A` command being unable to express a closed circle), and reversing that back into center/radius is genuinely ambiguous in the degenerate case — I verified this breaks for the "single 100%-share slice" scenario. `d3.arc().context(ctx)` (what `renderArc` already uses) sidesteps this entirely since it draws to the canvas context natively, without ever round-tripping through a path string — same pattern the Area/Line canvas rendering already uses in this codebase. So I kept that, just fixed the unsafe dispatch.

Pushed in bae38c7.

---

packages/react/src/lib/components/Plots/Pie/Donut/Donut.tsx
IPWright83: Agreed, that's cleaner. Inverted it: `<Donut>` is now the real component (own `DonutBase`, HOCs, full props including `innerRadius`, default `0.6`), and `<Pie>` is a thin wrapper that always renders `<Donut innerRadius={0}>` and doesn't expose `innerRadius` in its props at all — so it's not just a bad default anymore, it's not offered as an option.

Pushed in bae38c7.

---

packages/core/src/canvas/renderPath.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Renders a `<path>` element to the canvas. A `<path>` can represent many different shapes, so rather than
+ * assuming every path is the same kind of shape, this looks up the appropriate renderer using the node's
+ * `data-path-type` attribute
+ * @param  context             The Canvas context object to render to
+ * @param  node                The virtual DOM node that represents this element
+ * @param  overrideColor       A custom color to override the node color which is used for the virtual canvas
+ */
+export function renderPath(context: CanvasRenderingContext2D, node: Element, overrideColor?: IColor) {
```
IPWright83: Move this inline within renderElements and use a switch statement for clarity

packages/react/src/lib/components/Plots/Pie/Pie/Pie.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+import type { IDonutProps } from "../Donut";
+import { Donut } from "../Donut";
+
+export type IPieProps = Omit<IDonutProps, "innerRadius">;
+
+/**
+ * Represents a Pie plot. This is a `<Donut>` with no hole in the middle, so unlike `<Donut>` it doesn't
```
IPWright83: Remove the comment about inner radius here, it's not required

packages/react/src/lib/components/Plots/Pie/Pie/Pie.unit.tsx
```diff
@@ -0,0 +1,197 @@
...
+                data,
+            });
+
+            // Wait for the enter transition to settle so the snapshot deterministically
+            // captures the final arc geometry rather than a mid-animation frame
+            await wait();
+            expect(asFragment()).toMatchSnapshot();
+        });
+
+        it("should always have a zero innerRadius, even if forced", async () => {
```
IPWright83: Remove this test

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+type IPieHierarchyNode = d3.HierarchyRectangularNode<IPieHierarchyDatum>;
+
+/**
+ * Ensures that every combination of the `x`/`x2` fields is unique and logs a warning if not
+ * @param  data     The full dataset
+ * @param  x        The name of the inner ring category field
+ * @param  x2       The name of the outer ring subdivision field
+ */
+function ensureCombinationsAreUnique(data: IData, x: string, x2: string): void {
```
IPWright83: Pull out this hook with the other warning hooks

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
```diff
@@ -0,0 +1,372 @@
...
+ */
+function buildHierarchy(data: IData, x: string, x2: string, y: string, sort: boolean) {
+    const groupedByX = d3.group(data, (d) => `${d[x]}`);
+
+    const children: IPieHierarchyDatum[] = Array.from(groupedByX, ([key, rows]) => {
+        const total = d3.sum(rows, (row) => Number(row[y]) || 0);
+
+        return {
+            key,
+            // The parent ring represents an aggregate across multiple rows, so its synthetic datum
```
IPWright83: Explain this more with an example

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
```diff
@@ -0,0 +1,372 @@
...
+ * Builds a 2-level hierarchy (`x` then `x2`) from a flat dataset, and lays it out radially so that
+ * each `x` category occupies an angular span proportional to its total, subdivided by `x2` within it
+ * @param  data     The full dataset
+ * @param  x        The name of the inner ring category field
+ * @param  x2       The name of the outer ring subdivision field
+ * @param  y        The name of the value field
+ * @param  sort     Should the slices be sorted by value (descending)?
+ * @return          The root of the laid out hierarchy
+ */
+function buildHierarchy(data: IData, x: string, x2: string, y: string, sort: boolean) {
```
IPWright83: Move this into a separate file

packages/react/src/lib/components/Plots/Pie/useFocused.ts
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * The theme for the chart
+     */
+    theme: ITheme;
+}
+
+/**
+ * Helper function to highlight a selected slice on a Pie/Donut plot
+ * @return              A function to clear up the highlight
+ */
```
IPWright83: Put this inline instead

packages/react/src/lib/components/Plots/Pie/useTooltip.ts
```diff
@@ -0,0 +1,110 @@
...
+    name: string;
+    value: IValue;
+    color: IColor;
+}
+
+/**
+ * Helper function to manage tooltips for a selected slice on a Pie/Donut plot
+ * @return              A function to clear up the tooltip
+ */
+function tooltip({ dispatch, datum, name, value, color, event }: IPieTooltipProps) {
```
IPWright83: Move this code inline in the useEffect

packages/react/src/lib/hoc/withRadialPlot.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import React, { useEffect, useState } from "react";
+import { useSelector } from "react-redux";
+
+/**
+ * Wraps a polar plot (e.g. Pie/Donut) and handles some of the basics such as missing layers,
+ * along with calculating the center point & maximum radius available to the plot
+ * @param  WrappedComponent     The D3 layer to render
+ * @return                      The wrapped layer
+ */
```
IPWright83, line 12: Could we rename this, instead of polar, use the word radial

packages/react/src/lib/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { LegendOverlay } from "../LegendOverlay";
+import { TooltipOverlay } from "../TooltipOverlay";
+
+export type IPieChartProps = IChartProps;
+
+/**
+ * Represents a Pie/Donut chart. This is the polar equivalent of the `<XYChart>`, and is
+ * intended to wrap `<Pie>`, `<Donut>` or `<StackedDonut>` plots
+ */
+export const PieChart = forwardRef<IChartRef, IPieChartProps>(({ children, ...props }, ref) => {
```
IPWright83: Let's rename this to a Radial chat, which then contains a Pie or Donut. Otherwise we end up with Pie chart>Pie

packages/core/src/canvas/renderPath.ts
IPWright83: Done - inlined into `renderElements.ts`'s switch (nested switch on `data-path-type` within the `PATH` case), `renderPath.ts` removed. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/Pie/Pie.tsx
IPWright83: Done. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/Pie/Pie.unit.tsx
IPWright83: Removed. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
IPWright83: Moved to `@chart-io/core/utils/checks/ensureCombinationsAreUnique.ts`, alongside `ensureBandwidth`/`ensureValuesAreUnique`/`ensureNoScaleOverflow`. Generalized it to take a `fields: string[]` array rather than being hardcoded to `x`/`x2`, for consistency with how the other checks are written for reuse. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
IPWright83: Added a worked example (region/product/sales) to the docstring showing how the flat rows map to the 2-level hierarchy and resulting angles. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
IPWright83: Moved to `StackedDonut/buildHierarchy.ts`, mirroring the `StackedColumn/getParentKey.ts` precedent. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/useFocused.ts
IPWright83: Inlined into the `useEffect`. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/useTooltip.ts
IPWright83: Inlined into the `useEffect`. Pushed in f1fd872.

---

packages/react/src/lib/hoc/withRadialPlot.tsx
IPWright83: Renamed to `withRadialPlot` (file, export, all call sites and doc references). Pushed in f1fd872.

---

packages/react/src/lib/components/PieChart/PieChart.tsx
IPWright83: Renamed `PieChart` -> `RadialChart` (folder, component, exports, stories, docs), and the Storybook "PieCharts" grouping -> "RadialCharts" to match. Now reads as `<RadialChart><Pie/></RadialChart>` or `<RadialChart><Donut/></RadialChart>`. Pushed in f1fd872.

---

packages/react/src/lib/components/Plots/Pie/Donut/DonutBase.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+export interface IDonutBaseProps {
+    /**
+     * The layer to be rendered upon. Typically this is an `<svg:g>` or a fake HTMLElement when using canvas.
+     */
+    layer?: React.MutableRefObject<Element>;
+    /**
+     * The key of the field used for the category/label of each slice
+     */
+    x: string;
```
IPWright83: This should be called category, not x

packages/react/src/lib/components/Plots/Pie/Donut/DonutBase.tsx
```diff
@@ -0,0 +1,268 @@
...
+    /**
+     * The key of the field used for the value of each slice
+     */
+    y: string;
```
IPWright83: This should be called value, not y

packages/react/src/lib/components/Plots/Pie/StackedDonut/buildHierarchy.ts
```diff
@@ -0,0 +1,57 @@
...
+ * For example, given `x="region"`, `x2="product"`, `y="sales"` and the data
+ * `[{ region: "North", product: "Widgets", sales: 5 }, { region: "North", product: "Gadgets", sales: 5 },
+ * { region: "South", product: "Widgets", sales: 10 }]`, this builds a root with two depth-1 children
+ * ("North" with a synthetic `datum: { region: "North", sales: 10 }`, and "South" with
+ * `datum: { region: "South", sales: 10 }`), each with their own depth-2 children for "Widgets"/"Gadgets".
+ * "North" ends up spanning half the circle (angle `[0, π]`) since it accounts for half the total sales,
+ * split evenly between its two depth-2 children ("Widgets" getting `[0, π/2]`, "Gadgets" getting `[π/2, π]`)
+ * @param  data     The full dataset
+ * @param  x        The name of the inner ring category field
+ * @param  x2       The name of the outer ring subdivision field
```
IPWright83: X2 should be subCategory

packages/react/src/lib/components/Plots/Pie/Donut/DonutBase.tsx
IPWright83: Renamed `x` -> `category` across `<Pie>`/`<Donut>` (and `<StackedDonut>`, where it's the inner ring): `DonutBase.tsx`, `StackedDonutBase.tsx`, `buildHierarchy.ts`, plus stories, unit tests, and the `Pie.mdx`/`StackedDonut.mdx` prop tables. Left the unrelated `cx`/`cy` center-coordinate props and mouse-event `x`/`y` alone. Pushed in 0e30414.

---

packages/react/src/lib/components/Plots/Pie/Donut/DonutBase.tsx
IPWright83: Renamed `y` -> `value` in the same pass (all three components). Pushed in 0e30414.

---

packages/react/src/lib/components/Plots/Pie/StackedDonut/buildHierarchy.ts
IPWright83: Renamed `x2` -> `subCategory` on `<StackedDonut>`, including the W008 warning doc (was "duplicate x/x2 combination", now "duplicate category/subCategory combination"). Pushed in 0e30414.

---

IPWright83:
We want to support more stacks. I feel like this might be a different flavour of the Stacked pie that takes a hierarchical dataset. https://observablehq.com/@kerryrodden/sequences-sunburst

I also think it would be good to support central values like this instead of the traditional tooltip

PR #223, opened 2026-07-27
StackedDonut: N-level sunburst, center-value display, generic GDP data

IPWright83:
## Summary

Follow-up to #221, addressing the "stacked pies" feedback left on that PR:

- **N-level hierarchy**: `<StackedDonut>`'s `category`/`subCategory` props are replaced with a single `categories: string[]` prop (one field per ring, innermost first). `buildHierarchy` now groups recursively to arbitrary depth instead of a fixed 2 levels, and the ring radius calculation, recursive `colorFor`, and breadcrumb-based tooltip name all generalize to N rings. This is a proper sunburst now, similar in spirit to the [sequences sunburst](https://observablehq.com/@kerryrodden/sequences-sunburst) example referenced in the review comment.
- **Center value display**: `<RadialChart>` gains a `centerValue` prop. When set, a new `CenterValueOverlay` shows the hovered slice's breadcrumb name and formatted value in the center of the donut's hole, replacing the floating Tooltip - useful for deep hierarchies where a floating tooltip can obscure the chart.
- **Generic sample data**: Pie/Donut/StackedDonut stories now use a new `gdp_dataset` (continent/country/sector GDP figures) instead of the sales dataset's `Item Type` field, which surfaced values like "Baby Food".

## Test plan

- [x] `pnpm --filter @chart-io/react test` / `pnpm --filter @chart-io/core test` - all passing
- [x] `pnpm --filter @chart-io/react types` / `lint` - clean
- [x] `pnpm --filter @chart-io/core types` / `lint` - clean
- [x] Verified visually via Storybook + Playwright: Pie/Donut with GDP data, 2-ring and 3-ring StackedDonut sunbursts, and the center-value hover interaction (confirmed the floating tooltip is suppressed and the center text updates correctly)

---

### DIFF
--- a/packages/react/src/lib/components/RadialChart/RadialChart.tsx
+++ b/packages/react/src/lib/components/RadialChart/RadialChart.tsx
@@ -1,20 +1,28 @@
 import React, { forwardRef } from "react";
 
+import { CenterValueOverlay } from "../CenterValueOverlay";
 import { Chart, IChartProps, IChartRef } from "../Chart";
 import { LegendOverlay } from "../LegendOverlay";
 import { TooltipOverlay } from "../TooltipOverlay";
 
-export type IRadialChartProps = IChartProps;
+export interface IRadialChartProps extends IChartProps {
+    /**
+     * Displays the hovered datapoint's name/value in the center of the chart's hole (e.g. a `<Donut>`
+     * or `<StackedDonut>`) instead of in a floating Tooltip
+     * @default false
+     */
+    centerValue?: boolean;
+}
 
 /**
  * Represents a radial chart. This is the polar equivalent of the `<XYChart>`, and is
  * intended to wrap `<Pie>`, `<Donut>` or `<StackedDonut>` plots
  */
-export const RadialChart = forwardRef<IChartRef, IRadialChartProps>(({ children, ...props }, ref) => {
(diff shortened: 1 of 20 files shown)


packages/react/src/lib/components/RadialChart/RadialChart.tsx
```diff
@@ -1,20 +1,28 @@
...
 import { LegendOverlay } from "../LegendOverlay";
 import { TooltipOverlay } from "../TooltipOverlay";
 
-export type IRadialChartProps = IChartProps;
+export interface IRadialChartProps extends IChartProps {
+    /**
+     * Displays the hovered datapoint's name/value in the center of the chart's hole (e.g. a `<Donut>`
+     * or `<StackedDonut>`) instead of in a floating Tooltip
+     * @default false
+     */
```
IPWright83, line 15: Can we default this to true if showing donut plots?

IPWright83: Done in ccfdb00 - `RadialChart` now checks its children for a `Donut`/`StackedDonut` (via a `hasCenterHole` static flag) and defaults `centerValue` to `true` when one is present, `false` for `Pie` (no hole). Still overridable explicitly either way. Updated the stories/docs to show the new default plus an explicit `centerValue={false}` opt-out example.

---

PR #227, opened 2026-08-01
Add Radar and RadialArea charts (AngleAxis/RadialAxis, Canvas support)

IPWright83:
## Summary

Two new radial chart types:

- **`<Radar>`**: plots one closed, filled (or, via `filled={false}`, stroke-only) polygon per series, connecting a point per field in `ys` around a shared angular scale - one row of `data` per series. Each vertex gets a small marker to support hover/click tooltips. Supports per-spoke independent domains (e.g. a 1-5 rating next to a percentage), a `labeller` for mapping field keys to display labels, and multi-series comparisons.
- **`<RadialArea>` / `<RadialAreas>`**: a filled area from the center outward, following a continuous angular domain (e.g. a year of dates) around the circle - the polar equivalent of `<Area>`, matching the shape behind [Observable's radial area chart example](https://observablehq.com/@d3/radial-area-chart/2). Renders directly to Canvas via d3's shape-generator context API. Hover finds the nearest point by inverting the mouse position through the angular scale, mirroring `<Area>`'s own interaction model.

Supporting infrastructure, shared by both:

- **`<AngleAxis>` / `<RadialAxis>`** (plus matching `AngleScale`/`RadiusScale`): draw the spokes/category labels and concentric rings/value labels a radial plot needs to be readable. `<AngleAxis>` supports both categorical (`"point"`/`"band"`) and continuous (`"time"`/`"linear"`) scale types - continuous spokes are evenly-spaced ticks rather than one per data point. `<RadialAxis>` rings show real tick values when they resolve to a single scale (one field, or multiple sharing an explicit `domain`), falling back to a normalized percentage only when fields are independently scaled. The scale-rendering half (subscribes to the resolved scale) and the scale-registering half (dispatches it) are kept as separate sibling components - combining them caused a render/dispatch feedback loop, since resolving a scale always returns a fresh object.
- **Canvas support**: `@chart-io/core` gains a `renderPolygon` primitive (plus a `POLYGON` case in `renderElements`'s dispatch) so Radar's canvas mode renders and hit-tests the same way SVG mode does. RadialArea instead draws straight to the canvas context (like `<Area>` already does), so no new primitive was needed there.
- **`<RadialChart>`** now includes an `<EventReceiver>` and `<Markers>` overlay (matching `<XYChart>`), so radial plots can support hover driven by global mouse position, not just per-element hover.
- `createLabeller` (`@chart-io/core`): maps a field key (e.g. `"avg_score"`) to a display label (e.g. `"Mean Score"`), used by `<Radar>`'s tooltip/legend/axis labels - a first step towards future translation/formatting support.
- Moved `useTooltip`/`useFocused` from `Plots/Pie/` up to the shared `Plots/` level so other plots can reuse the same tooltip/focus wiring as `Donut`/`StackedDonut`.

## Test plan

- [x] `pnpm --filter @chart-io/react test` / `pnpm --filter @chart-io/core test` - all passing
- [x] `pnpm --filter @chart-io/react types` / `lint` - clean
- [x] `pnpm --filter @chart-io/core types` / `lint` - clean
- [x] Verified visually via Storybook + Playwright: SVG and Canvas rendering (including hover/tooltip interactions), single- and multi-series, for both Radar and RadialArea

---


packages/core/src/canvas/renderPolygon.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Renders a closed polygon to the canvas
+ * @param  context             The Canvas context object to render to
+ * @param  node                The virtual DOM node that represents this element
+ * @param  overrideColor       A custom color to override the node color which is used for the virtual canvas
+ */
+export function renderPolygon(context: CanvasRenderingContext2D, node: Element, overrideColor?: IColor) {
+    const selection = d3.select(node);
+    const points = parsePoints(selection.attr("points") ?? "");
```
IPWright83, line 28: Is this a reliable way of converting to canvas?

packages/react/src/lib/components/Axis/AngleAxis/AngleAxis.tsx
```diff
@@ -0,0 +1,145 @@
...
+    tickFormat?: (value: IValue) => string;
+}
+
+/**
+ * Renders the spokes and category labels for an AngleAxis. Kept separate from `<AngleAxis>` itself
+ * since this subscribes to the resolved scale (and so re-renders whenever it's set), whereas the
+ * sibling `<AngleScale>` that sets it must not, to avoid the two re-triggering one another
+ * @return The visual part of the AngleAxis
+ */
+function AngleAxisRenderer({
```
IPWright83: Can we call this an AxisSpoke and move it below the other component in the file?

packages/react/src/lib/components/Axis/RadiusAxis/RadiusAxis.tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * Renders the concentric rings and value labels for a RadiusAxis. Kept separate from
+ * `<RadiusAxis>` itself since this subscribes to the resolved scale (and so re-renders whenever
+ * it's set), whereas the sibling `<RadiusScale>` that sets it must not, to avoid the two
+ * re-triggering one another
+ * @return The visual part of the RadiusAxis
+ */
+function RadiusAxisRenderer({
```
IPWright83: Let's call this a RadialAxisRing and move below the other component. Rename the main component and file RadialAxis.

IPWright83: Does this cope with each axis potentially having a different domain? E.g. 1-5 on 1 axis but a percentage on another? If not we should make it and add example stories

packages/react/src/lib/components/Plots/Radar/RadarSeries/RadarSeriesBase.tsx
```diff
@@ -0,0 +1,245 @@
...
+     */
+    layer?: React.MutableRefObject<Element>;
+    /**
+     * The key of the field used for the angular (category) position of each point
+     */
+    category: string;
+    /**
+     * The key of the field used for the radial (value) position of each point
+     */
+    y: string;
```
IPWright83: Should this not be ys? That then enables us to have multiple series on a radar chart

packages/react/src/lib/components/Plots/Radar/Radar.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+        theme,
+        leftMargin: margin,
+        rightMargin: margin,
+        topMargin: margin,
+        bottomMargin: margin,
+    },
+} as Meta<typeof Radar>;
+
+const data = [
+    { skill: "Speed", playerA: 80, playerB: 60 },
```
IPWright83: Could we instead have a single row per series?

packages/core/src/canvas/renderPolygon.ts
IPWright83: Yes - it's safe here because a `<polygon>`'s `points` attribute is a plain, unambiguous list of comma/space-separated numbers (`"x1,y1 x2,y2 ..."`), unlike an SVG path `d` string which can contain curves/arcs and is lossy to round-trip. Splitting on whitespace and pairing coordinates into `moveTo`/`lineTo`/`closePath` calls is a complete, deterministic mapping with no edge cases to worry about (no relative commands, no curve approximation). This is different from the arc-serialization case we hit earlier in the Donut work, where the path string itself needed reconstructing - `points` doesn't have that problem.

---

packages/react/src/lib/components/Axis/AngleAxis/AngleAxis.tsx
IPWright83: Done - renamed to `AxisSpoke` and moved it below the public `AngleAxis` component in the file.

---

packages/react/src/lib/components/Axis/RadiusAxis/RadiusAxis.tsx
IPWright83: Done - renamed the file/component to `RadialAxis`, and the internal ring renderer to `RadialAxisRing`, moved below the public component.

---

IPWright83: Good catch - it didn't. Reworked `RadiusScale` so each field gets its own independently-computed d3 scale (own domain, auto-typed so a string/boolean field naturally gets a band/point scale) all sharing the same `[0, maxRadius]` pixel range, instead of one shared scale across all fields. `RadialAxisRing` no longer reads any single field's scale at all - since there's no longer one "real" domain to represent, it always draws normalized rings on a fixed `[0, 1]` scale (labelled as a percentage by default, overridable via `tickFormat`). Added a "Different Domains per Spoke" story with a GPU-comparison dataset (memory in GB, CUDA core count, TFLOPS, a 1-10 score) to demonstrate it - each spoke keeps its own scale but they land on the same normalized rings so they stay visually comparable. If you do want every field to share one literal domain, `<RadialAxis aggregate>` is still available as an escape hatch.

---

packages/react/src/lib/components/Plots/Radar/RadarSeries/RadarSeriesBase.tsx
IPWright83: Agreed - reworked the data model so `data` has one row per series (see reply below), and `RadarSeriesBase` now takes a `ys: string[]` prop: the ordered list of fields to plot, one spoke per field, read from that series' own row. `<Radar>` then renders one `<RadarSeries>` per row of `data`, so multiple series "just work". Added a `filled` prop too (stroke-only outline when `false`), which is useful when overlaying many series so the fills don't obscure one another - see the new "Unfilled (Overlapping Trends)" story.

---

packages/react/src/lib/components/Plots/Radar/Radar.stories.tsx
IPWright83: Done - `data` is now one row per series (e.g. `{ player: "Player A", Speed: 80, Power: 65, ... }`), with `<Radar name="player" ys={skills}>` rendering one `<RadarSeries>` per row. Also added `ThreeSeries` and `Unfilled` (5 series) stories to cover multi-series more explicitly.

---

packages/react/src/lib/components/Plots/Radar/Radar.stories.tsx
```diff
@@ -102,7 +114,78 @@ export const SingleSeries = {
...
+        animationDuration={args.animationDuration}
+        theme={args.theme}
+        useCanvas={args.useCanvas}
+        onClick={args.onClick}
+        onMouseOver={args.onMouseOver}
+        onMouseOut={args.onMouseOut}
+    >
+        <AngleAxis fields="category" domain={gpuFields} tickFormat={gpuLabeller} />
+        <RadialAxis fields={gpuFields} />
+        <Radar category="category" name="gpu" ys={gpuFields} labeller={gpuLabeller} />
```
IPWright83: Is category used?

packages/react/src/lib/components/Plots/Radar/Radar.stories.tsx
```diff
@@ -102,7 +114,78 @@ export const SingleSeries = {
...
+    </RadialChart>
```
IPWright83, line 192: Can the labeller be global to a chart, rather than just a Radar? XYCharts and Pies should support it too, so it needs to be passed in at the chart level

packages/react/src/lib/components/Plots/Radar/Radar.stories.tsx
IPWright83: Yes - `category` is the store key `RadarSeriesBase` looks up its angle scale under (`chartSelectors.scales.getScale(state, category, "plot")`), so it has to match the sibling `<AngleAxis>`'s `fields`. Not dead code, but you're right that requiring it on every `<Radar>` was needless ceremony for the common case - it's just an arbitrary key, not a real data column. Made it optional with a default of `"category"` (matching `<AngleAxis>`'s own common default), so it only needs to be set explicitly when a chart has more than one independent Radar - e.g. two separate `<Radar>`/`<AngleAxis>` pairs on the same `<RadialChart>`, which would otherwise register their scales under the same key and clobber one another. Documented that case with an example in the MDX docs.

---

packages/react/src/lib/components/Plots/Radar/Radar.stories.tsx
IPWright83: Agreed - moved it to the chart level, mirroring exactly how `theme` already works: `<Chart labeller>` dispatches into the redux store (`chartActions.setLabeller`, new `chartSelectors.labeller`), and any descendant plot reads it back via the selector rather than through prop drilling. So it's now available to `RadialChart`/`XYChart`/any chart type for free - I haven't retrofitted every existing plot type (Line/Bar/Pie/etc.) to actually consume it yet, only Radar does so far, but the chart-level plumbing itself isn't Radar-specific anymore, so that's a small follow-up per plot type rather than a redesign. `<Radar>` (and each `RadarSeries`) still accepts its own `labeller` prop too, which overrides the chart-level one for just that series if needed. Updated the `DifferentDomains` story to set it via `<RadialChart labeller={gpuLabeller}>` instead of on `<Radar>` directly.

---

packages/react/src/lib/components/Scale/RadiusScale.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { AutoScale, IAutoScaleProps } from "./AutoScale";
+import { useArray } from "../../hooks";
+
+export type IRadiusScaleProps = Omit<IAutoScaleProps, "range">;
+
+/**
+ * Represents a RadiusScale, mapping value field(s) to a radius (in pixels) from the center of a
+ * circle. Used by radial plots such as `<Radar>`/`<RadialArea>` and their `<RadialAxis>`.
+ *
+ * Unless `aggregate` is set, each field is given its own independently computed domain (e.g. a
```
IPWright83: There should be an explanation somewhere on the use case for aggregate

packages/react/src/lib/components/Scale/RadiusScale.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * this is what lets each spoke of a `<Radar>` represent a different domain. To instead compare
+ * multiple fields on one common domain (e.g. multiple `<RadialArea>` series in the same units),
+ * pass an explicit `domain` override - every field will then be scaled against that same domain
+ * @param  props   Props for the scale
+ * @return         A scale component
+ */
+export function RadiusScale({ fields, scaleType, domain, aggregate = false }: IRadiusScaleProps) {
+    const plotWidth = useSelector((s: IState) => chartSelectors.dimensions.plot.width(s));
+    const plotHeight = useSelector((s: IState) => chartSelectors.dimensions.plot.height(s));
+    const maxRadius = Math.max(0, Math.min(plotWidth, plotHeight) / 2);
```
IPWright83: I think we calculate this in a few places? Can we use a selector instead?

packages/react/src/lib/components/Plots/RadialArea/RadialAreas.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return (
+        <React.Fragment>
+            {ys.map((y, i) => (
+                <RadialArea {...props} key={y} y={y} color={palette[i % palette.length]} />
+            ))}
+        </React.Fragment>
+    );
+}
+
+RadialAreas.requiresVirtualCanvas = false;
```
IPWright83, line 51: Does this not require a virtual canvas?

packages/react/src/lib/components/Plots/RadialArea/RadialArea.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+        height={args.height}
+        animationDuration={args.animationDuration}
+        theme={args.theme}
+        useCanvas={args.useCanvas}
+        onClick={args.onClick}
+        onMouseOver={args.onMouseOver}
+        onMouseOut={args.onMouseOut}
+    >
+        <AngleAxis fields="date" scaleType="time" ticks={12} tickFormat={(d) => (d as Date).toLocaleString("en-US", { month: "short" })} />
+        {/* An explicit domain puts both series on the same scale, so their absolute temperatures stay comparable */}
```
IPWright83: Put this in th docs on the story, I believe a comment is sufficient with autodocs?

packages/react/src/lib/components/Plots/RadialArea/RadialArea.mdx
```diff
@@ -0,0 +1,76 @@
+import { Story, Canvas, Meta } from "@storybook/blocks";
+import * as RadialAreaStories from "./RadialArea.stories"
+
+<Meta title="RadialCharts/RadialArea" />
+
+# RadialArea Plots
+
+A `<RadialArea>` fills the area from the center outward to a value, following a continuous angular
+domain around the circle - the polar equivalent of `<Area>`, and the shape behind
```
IPWright83: Remove the reference Observable

packages/react/src/lib/components/Plots/RadialArea/RadialArea.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+        height,
+        theme,
+        leftMargin: margin,
+        rightMargin: margin,
+        topMargin: margin,
+        bottomMargin: margin,
+    },
+} as Meta<typeof RadialArea>;
+
+// A year of daily temperatures, wrapped around a full circle by day of year - the classic
```
IPWright83: Remove the reference to Observable

packages/react/src/lib/components/Plots/RadialArea/RadialArea/RadialAreaBase.tsx
```diff
@@ -0,0 +1,172 @@
...
+     */
+    canvas?: HTMLCanvasElement;
+    /**
+     * An internal setting, to prefer using a brush scale. You should not need to use it directly
+     */
+    scaleMode?: IScaleMode;
+}
+
+/**
+ * Represents a RadialArea plot: a single filled area from the center outward to `y`, following a
```
IPWright83, line 79: Do we need a separate component to the radar for this? Isn't this just the radar with a single scale for rotation?

packages/react/src/lib/components/Scale/RadiusScale.tsx
IPWright83: Added - the JSDoc now explains what `aggregate` actually does (sums every field's value per row into one domain, same as `<YAxis aggregate>` for a stacked chart's total) and when you'd reach for it vs. the default vs. an explicit `domain` override. Also fixed a related inaccuracy: the old wording implied `aggregate` gives a "shared domain unchanged", but it's summed - if you want fields to share one domain unchanged (not summed), that's what `domain` is for. Corrected the same phrasing in `RadialAxis.tsx` and `Radar.mdx` too.

---

packages/react/src/lib/components/Scale/RadiusScale.tsx
IPWright83: Good catch - it was duplicated in `RadiusScale`, `AngleAxis`, `RadialAxis`, `withRadialPlot`, and `CenterValueOverlay` (5 places, all computing the same `left + width/2` / `top + height/2` / `min(width,height)/2`). Added `chartSelectors.dimensions.plot.cx`/`cy`/`maxRadius` and switched all 5 over to it.

---

packages/react/src/lib/components/Plots/RadialArea/RadialAreas.tsx
IPWright83: No - it mirrors `Area`/`Lines`/`StackedArea` (all `requiresVirtualCanvas = false` too), not `Radar`/`Donut`. `RadialAreaBase` draws the shape directly to the real canvas context via d3's shape-generator `.context()` API (same technique `AreaBase` already uses), and finds the hovered point by inverting the global mouse position through the angular scale rather than per-pixel colour hit-testing - so there's no need for the virtual canvas's color-picking layer. Added a comment above both `requiresVirtualCanvas` flags explaining this, since it's easy to assume every radial plot needs it like Radar does.

---

packages/react/src/lib/components/Plots/RadialArea/RadialArea.stories.tsx
IPWright83: Moved - it's already covered in `RadialArea.mdx`'s "Multiple Series" section, so the inline JSX comment was just duplicating that. Replaced it with a plain `//` comment above the template pointing at the docs, so the explanation lives in one place.

---

packages/react/src/lib/components/Plots/RadialArea/RadialArea.mdx
IPWright83: Removed.

---

packages/react/src/lib/components/Plots/RadialArea/RadialArea.stories.tsx
IPWright83: Removed.

---

packages/react/src/lib/components/Plots/RadialArea/RadialArea/RadialAreaBase.tsx
IPWright83: Fair question - they do share the angular/radial axis infrastructure (`AngleAxis`/`RadialAxis`/`AngleScale`/`RadiusScale`), and I did consider folding RadialArea into Radar since "a Radar with a continuous scale" is a reasonable mental model. I kept them separate for the same reason `<Area>` and `<Line>` are separate components rather than "a Line with fill", even though they also share `XAxis`/`YAxis`:

- **Data shape**: Radar reads one row per series and iterates over *fields* within that row (`ys`) to build each vertex. RadialArea reads one field for *many* rows (typically hundreds, e.g. 365 days) to build a continuous path. These are opposite iteration axes over the data, not a parameterization of the same loop.
- **Interactivity**: Radar has a handful of vertices, so it renders a real marker per vertex with its own mouseover/mouseout/click handlers (and needs the virtual canvas for Canvas-mode hit-testing). RadialArea has too many points for that to be practical, so it finds the nearest point from the global mouse position via `xScale.invert()` + bisection - the same interaction model `Area`/`Line` already use, and why it *doesn't* need the virtual canvas (see the other comment thread).
- **Rendering/animation**: Radar's shape is a small, fixed-size array of points animated via a custom point-interpolation tween. RadialArea's shape is an SVG path (or direct canvas draw) built by d3's `areaRadial` generator and animated via `interpolateMultiPath` - the same path-interpolation `Area`/`Line` use.

So the two are structurally closer to "the polar Area" and "the polar version of a marker-based comparison chart" than variations of one component - merging them would mean branching most of the internals (data iteration, interaction model, render/animation strategy) on a flag, which felt like it'd make one component do two genuinely different jobs rather than simplify anything. Happy to revisit if you see a cleaner unification though.

---

IPWright83: Is this clear/transparent enough for an end user?

IPWright83: Wonder if we should call the RadialArea a Polar chart instead?

IPWright83: Fair - that explanation was PR-comment-only, not something an actual user would ever see. Added a proper "`<RadialArea>` vs `<Radar>`" section near the top of `RadialArea.mdx` in plain terms (fixed named categories → Radar; continuous measurement over a cycle → RadialArea), plus a one-line pointer the other way at the top of `Radar.mdx` so someone landing on either page finds the right one.

---

IPWright83: I'd lean towards keeping `RadialArea`, for two reasons:

1. **Naming collision**: "Polar Area Chart" already means something specific and different in dataviz - the Nightingale rose/coxcomb diagram, where each category gets a fixed angular wedge and *varying radius* (basically a Pie chart with variable slice length). That's much closer to what `<Radar>` or even `<Pie>` does than what we've built here (a continuous area following a continuous angle, more like an Area/Line chart wrapped into polar coordinates). Calling ours "Polar" risks users expecting the wedge-chart and being confused when they get a continuous area instead.
2. **Consistency**: everything else in this family is named "Radial" - `RadialChart`, `RadialAxis`, `withRadialPlot`. "PolarArea" would be the one outlier.

That said, it's a genuine judgment call and I don't feel strongly - if you'd still rather have `PolarArea` after that, say the word and I'll do the rename (component, file, folder, stories, docs, tests). Leaving it as `RadialArea` for now.

---

PR #229, opened 2026-08-09
Move buildHierarchy into @chart-io/core, decoupled from layout

IPWright83:
## Summary

First of a planned series of PRs building towards a treemap, dendrogram and radial dendrogram chart, sharing a hierarchical data structure with `<StackedDonut>`.

- Moves the hierarchy-building logic behind `<StackedDonut>` out of `packages/react` and into `@chart-io/core` as a shared `buildHierarchy` utility.
- Decouples it from any particular layout: it now only groups, sums and (optionally) sorts a flat dataset into a `d3.hierarchy`, leaving the caller to apply its own layout on top. `<StackedDonut>` now applies `d3.partition()` itself, immediately after calling `buildHierarchy`.
- This sets up the upcoming chart types to reuse the same builder with their own layout (`d3.treemap()`, `d3.cluster()`/`d3.tree()`, and a polar-mapped variant for the radial dendrogram) instead of duplicating the grouping/summing logic.
- Negative values in the hierarchy's `value` field are now clamped to `0` (none of these layouts - an angular span, a rectangle's area, tree connectivity - can represent a negative size) and log a new `W009` warning, rather than silently producing a broken or degenerate layout as before.

This PR is purely a refactor plus the negative-value handling - no new chart types or visible behavior changes for `<StackedDonut>` (all existing snapshots pass unchanged).

## Test plan

- [x] `pnpm --filter @chart-io/core test` - 13/13 suites pass, including new `buildHierarchy` unit tests
- [x] `pnpm --filter @chart-io/react test` - 69/69 suites pass (290 tests), all `StackedDonut` snapshots unchanged
- [x] Added a changeset (`@chart-io/core` minor, `@chart-io/react` patch)

---


packages/core/src/utils/hierarchy/buildHierarchy.unit.ts
```diff
@@ -0,0 +1,89 @@
...
+            { region: "South", sales: 10 },
+        ];
+
+        const root = buildHierarchy(negativeData, ["region"], "sales", false, "unit_test");
+
+        const [north, south] = root.children ?? [];
+        expect(north.value).toBe(0);
+        expect(south.value).toBe(10);
+
+        expect(spy.mock.calls[0][0]).toMatchSnapshot();
```
IPWright83: Don't use a snapshot for this assertion

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
```diff
@@ -152,7 +153,8 @@ export function StackedDonutBase({
     useRender(() => {
         ensureCombinationsAreUnique(data, categories, "StackedDonut");
 
-        const root = buildHierarchy(data, categories, value, sort);
+        const hierarchy = buildHierarchy(data, categories, value, sort, "StackedDonut");
```
IPWright83, line 171: We should allow injection of this function into charts in case the data has some wacky structure

PR #232, opened 2026-08-09
Add a Treemap chart

IPWright83:
## Summary

Second of the planned hierarchical chart PRs (after #229, the shared `buildHierarchy` refactor). Adds `<Treemap>`, the rectangular counterpart to `<StackedDonut>`.

- `<Treemap>` subdivides the plot area into nested rectangles from a flat dataset: `categories` is an ordered list of fields, outermost group first, and each leaf cell's area is proportional to `value`. It reuses `@chart-io/core`'s `buildHierarchy` (grouping/summing/sorting) and applies `d3.treemap()` on top, mirroring how `<StackedDonut>` applies `d3.partition()`.
- Only leaf cells are drawn as `<rect>`s - the grouping above them stays visually apparent from `d3.treemap()`'s own nested padding, without needing separate rectangles per group.
- Colors reuse `<StackedDonut>`'s exact pattern: each top-level category gets a color from the theme/palette, and deeper cells inherit a brightened version of their parent's color.
- Negative values are handled identically to `<StackedDonut>` (clamped to `0`, `W009` warning) since both go through the same shared `buildHierarchy`.
- Also accepts a `buildHierarchy` override prop (mirroring the one just added to `<StackedDonut>` in #229) for datasets that don't fit the flat group-by-`categories` shape.
- Since `<Treemap>` lays itself out within the full plot area rather than against x/y scales, it doesn't fit `<XYChart>` (needs axes/scales) or `<RadialChart>` (polar-specific). Added a new `<RectangularChart>` container (`Chart` + `TooltipOverlay` + `LegendOverlay`, no `EventReceiver`/`Markers` since tooltips are triggered per-cell rather than via nearest-point matching) and a `withRectangularPlot` HOC (injects the plot area's `left`/`top`/`width`/`height`, mirroring `withRadialPlot`'s `cx`/`cy`/`maxRadius`). Both are reusable by the upcoming Dendrogram chart, which needs the same rectangular layout area.
- Canvas mode needed no new core rendering primitives - `<rect>` elements are already dispatched generically by `renderElements`/`renderRect`, unlike `<StackedDonut>`'s arcs which needed a `data-path-type` dispatch.

Manually verified (via a static Storybook build + headless Chromium) that all 4 stories render correctly - SVG, Canvas, 3-level and 4-level hierarchies - including a real (non-synthetic) mouse hover on the Canvas story correctly hit-testing a cell and showing its breadcrumb/value tooltip.

## Test plan

- [x] `pnpm --filter @chart-io/core test` - 13/13 suites pass
- [x] `pnpm --filter @chart-io/react test` - 71/71 suites pass (299 tests), including new `Treemap` and `RectangularChart` unit tests
- [x] `pnpm build-storybook`, then manually verified all 4 Treemap stories render correctly in headless Chromium (screenshots + real mouse hover hit-testing on the Canvas story)
- [x] Added a changeset (`@chart-io/react` minor)

---


packages/core/src/utils/hierarchy/colorHierarchyNode.ts
```diff
@@ -0,0 +1,31 @@
+import { d3 } from "../../d3";
+
+import type { IHierarchyNode } from "./buildHierarchy";
+
+// Caps how light a brightened color can get, so a node with many siblings never washes out to
+// solid white (indistinguishable from the page background) against a light theme
+const MAX_LIGHTNESS = 0.85;
```
IPWright83, line 8: Can we go the other way too for dark theme

packages/react/src/lib/components/Plots/Treemap/TreemapBase.tsx
```diff
@@ -0,0 +1,268 @@
...
+     */
+    canvas?: HTMLCanvasElement;
+    /**
+     * This is an internally used function to allow the plot to render to a virtual canvas
+     */
+    renderVirtualCanvas?: (update: d3.Transition<Element, unknown, any, unknown>) => void;
+    /**
+     * The x-coordinate of the left edge of the plot area. Provided by `withRectangularPlot`
+     */
+    plotLeft?: number;
```
IPWright83: I don't like these, surely we can work this out from the existing selectors?

packages/react/src/lib/components/RectangularChart/RectangularChart.tsx
```diff
@@ -0,0 +1,26 @@
...
+export type IRectangularChartProps = IChartProps;
+
+/**
+ * Represents a rectangular chart, intended to wrap plots that lay themselves out within the full plot
+ * area rather than against x/y scales (e.g. `<Treemap>`), so don't need axes or scales. Tooltips on
+ * these plots are triggered directly per-element rather than by matching the nearest data point to the
+ * mouse position, so unlike `<XYChart>`/`<RadialChart>` this doesn't include an `<EventReceiver>` or
+ * `<Markers>` overlay
+ */
+export const RectangularChart = forwardRef<IChartRef, IRectangularChartProps>(({ children, ...props }, ref) => {
```
IPWright83: Can we remove this component? The tree map only really has a single plot so I think we can have this component be a Treemap and have an internal TreemapPlot that is added automatically. That'll mean adding some extra props to this component obviously.

packages/react/src/lib/hoc/withRectangularPlot.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * @param  WrappedComponent     The D3 layer to render
+ * @return                      The wrapped layer
+ */
+const withRectangularPlot = <P extends object>(WrappedComponent: React.ComponentType<P>) => {
+    /**
+     * Wraps a component and handles some of the potentially missing parts
+     * @param  {...any}    props        The rest of the props
+     * @return {ReactDOMComponent}      The wrapped layer
+     */
+    return function withRectangularPlot(props) {
```
IPWright83: I think this whole hoc is unnecessary

packages/react/src/lib/components/Plots/Treemap.mdx
```diff
@@ -0,0 +1,81 @@
+import { Story, Canvas, Meta } from "@storybook/blocks";
```
IPWright83, line 1: Make sure we update this with any changes

packages/core/src/utils/hierarchy/colorHierarchyNode.ts
IPWright83: Good call. `colorHierarchyNode` now takes the theme's background color and picks a direction based on it - brightens against a light background (unchanged), darkens against a dark one - both capped so a color can never wash out to solid white/black however many siblings a node has. Covered by two new regression tests.

---

packages/react/src/lib/components/Plots/Treemap/TreemapBase.tsx
IPWright83: Agreed - `chartSelectors.dimensions.plot.left/top/width/height` already existed. `TreemapBase` now pulls them directly via `useSelector`, the same way it already does for `width`/`height`/`theme`, instead of receiving them as HOC-injected props.

---

packages/react/src/lib/components/RectangularChart/RectangularChart.tsx
IPWright83: Went with your suggested shape: the old `Treemap` (the plot layer) is now `TreemapPlot`, and `Treemap` is a new self-contained component that renders `<RectangularChart><TreemapPlot/></RectangularChart>` internally, taking chart-level props (`data`/`width`/`height`/`zoomable`/etc) directly - no wrapper needed from consumers. Kept `RectangularChart` itself around rather than deleting it, since it's genuinely reusable plumbing (it'll do the same job for Dendrogram/RadialDendrogram/CirclePacking), but nobody has to reach for it directly to use `<Treemap>` anymore.

---

packages/react/src/lib/hoc/withRectangularPlot.tsx
IPWright83: Removed it entirely - `TreemapBase` now sources its plot dimensions straight from `chartSelectors.dimensions.plot.*`, so there was nothing left for the HOC to do.

---

packages/react/src/lib/components/Plots/Treemap.mdx
IPWright83: Updated - docs now reflect the self-contained `<Treemap>` (no more `<RectangularChart>` wrapper example), the chart-level props it accepts directly, and a new Zooming section covering `zoomable`/`breadcrumb`.

---

packages/react/src/lib/components/RectangularChart/RectangularChart.tsx
```diff
@@ -13,14 +21,17 @@ export type IRectangularChartProps = IChartProps;
...
-export const RectangularChart = forwardRef<IChartRef, IRectangularChartProps>(({ children, ...props }, ref) => {
-    return (
-        <Chart ref={ref} {...props}>
-            {children}
-            <TooltipOverlay onlyNearest={true} />
-            <LegendOverlay />
-        </Chart>
-    );
-});
+export const RectangularChart = forwardRef<IChartRef, IRectangularChartProps>(
```
IPWright83: I wanted to remove this component completely... Even if we have a bit of duplication. The contents of this should be in the highest Treemap component

packages/react/src/lib/components/Chart/Chart.tsx
```diff
@@ -154,6 +161,10 @@ export const Chart = forwardRef<IChartRef, IChartBaseProps>((props, ref) => {
         }
     }, [store.dispatch, labeller]);
 
+    useEffect(() => {
+        store.dispatch(chartActions.setZoomable(zoomable));
```
IPWright83: Does this make sense being in the chart overall? I wonder if it should just live in the Treemap?

packages/react/src/lib/components/RectangularChart/RectangularChart.tsx
IPWright83: Done - `RectangularChart` is deleted entirely. `Treemap` now renders `<Chart>` + `<TooltipOverlay>`/`<LegendOverlay>`/`<ZoomBreadcrumb>` directly itself, accepting the duplication. (Dendrogram/RadialDendrogram/CirclePacking will get the same treatment as a follow-up, since they currently share the same pattern.)

---

packages/react/src/lib/components/Chart/Chart.tsx
IPWright83: Agreed, moved it out. `zoomable` is no longer a `Chart` prop at all - `Chart` doesn't know "zoom" exists anymore. Since only descendants of `<Chart>` have access to the store it creates, `Treemap` now renders a small `<SetZoomable zoomable={zoomable} />` child (alongside `TreemapPlot`/`TooltipOverlay`/etc) that dispatches it - the same mechanism `<ZoomBreadcrumb>` already used for reading zoom state, just for writing the flag.

---

packages/react/src/lib/components/Plots/Treemap/Treemap.tsx
```diff
@@ -30,12 +45,15 @@ export const Treemap = forwardRef<IChartRef, ITreemapProps>(
...
+            zoomable = false,
+            breadcrumb = false,
             ...chartProps
         },
         ref,
     ) => {
         return (
-            <RectangularChart ref={ref} {...chartProps}>
+            <Chart ref={ref} {...chartProps}>
+                <SetZoomable zoomable={zoomable} />
```
IPWright83: Can we just dispatch to the store instead? If we can't due the store not being available then move the dispatch into the plot and pass zoom able down

IPWright83: Confirmed the store isn't available here - `Treemap.tsx` renders before `Chart`'s `withStore` HOC creates the `Provider`, so it can't dispatch directly. Went with your fallback: removed `SetZoomable` and moved the dispatch into `useZoom()` itself (it now optionally accepts `zoomable` and dispatches it as a side effect), with `TreemapBase` calling `useZoom(zoomable)` using its own `zoomable` prop passed down from `Treemap`. Pushed in 799f520.

---

packages/react/src/lib/components/Plots/Treemap/TreemapBase.tsx
```diff
@@ -110,6 +115,7 @@ export function TreemapBase({
     buildHierarchy = defaultBuildHierarchy,
     showInLegend = true,
     interactive = true,
+    zoomable = false,
```
IPWright83, line 118: Default to true

PR #233, opened 2026-08-09
Add Dendrogram and RadialDendrogram charts with zoom support

IPWright83:
## Summary

Third of the planned hierarchical chart PRs (after #229 the shared `buildHierarchy` refactor, and #232 the Treemap chart, which this branches from). Adds `<Dendrogram>` (linear) and `<RadialDendrogram>` (polar), plus a new zoom mechanism reusable by future hierarchical charts.

### Zoom infrastructure

- New chart-level `zoomable` prop (on `<RectangularChart>`/`<RadialChart>`, threaded through the same redux store pattern `labeller` already uses) - when enabled, clicking a node zooms in and reframes the layout on its subtree; clicking the current top-level node zooms back out one level.
- New optional `breadcrumb` prop, showing the current zoom path as a clickable trail ("All / North America / United States") for jumping straight to any ancestor.
- New `useZoom` hook and `<ZoomBreadcrumb>` overlay component, both reusable by the upcoming Circle Packing chart.

### `<Dendrogram>` / `<RadialDendrogram>`

- Both build on the same shared `buildHierarchy`/`colorHierarchyNode` utilities as `<StackedDonut>`/`<Treemap>`, applying `d3.cluster()` for the point layout - `<Dendrogram>` maps it to left-to-right pixel coordinates, `<RadialDendrogram>` maps the same x/y directly to angle/radius.
- Every node is drawn (unlike `<Treemap>`, which only draws leaves), connected by curved links, with an optional text label per node.
- Canvas mode needed two new core primitives: `renderLink` (a `linkHorizontal` cubic Bezier, replayed from `data-x0/y0/x1/y1` attributes) and `renderLinkRadial` (a `linkRadial` curve, replayed through `d3.linkRadial().context(...)` the same way `renderArc` already replays `d3.arc()`). Links and node circles are combined into one heterogeneous D3 join (mirroring `<Radar>`'s shape+marker pattern) so Canvas rendering has a single Transition covering both element types.

Manually verified via a static Storybook build + headless Chromium: SVG/Canvas rendering, N-level hierarchies, and the full zoom + breadcrumb interaction (clicking a node reframes the chart, breadcrumb updates and is clickable) for both charts. Also caught and fixed two real bugs along the way:
- Links were rendering as solid black filled wedges instead of thin stroked curves (a `.style("fill", null)` on the merged join was wiping out the `fill: none` set at enter).
- The Canvas hit-testing layer was rendering completely blank - fixed by combining links and nodes into one heterogeneous join instead of two separate `renderCanvas` calls, since the main canvas clears between calls.

**Known limitation**: `<RadialDendrogram>`'s Canvas story ships without a `play()` interaction test. The underlying Canvas hit-testing is confirmed working (unit tests pass, direct pixel/hit-color inspection checks out, and genuine Playwright mouse interaction succeeds), but the single-dispatch synthetic event `createCanvasTest` relies on doesn't reliably land on this chart's small (4px radius) node circles the way it does on other charts' larger shapes. Worth a follow-up if this pattern recurs elsewhere.

## Test plan

- [x] `pnpm --filter @chart-io/core test` - 14/14 suites pass
- [x] `pnpm --filter @chart-io/react test` - 74/74 suites pass (320 tests), including new zoom/breadcrumb, `Dendrogram` and `RadialDendrogram` unit tests
- [x] `pnpm build-storybook`, then manually verified all stories (SVG, Canvas, N-level, zoomable+breadcrumb) render correctly in headless Chromium for both charts
- [x] Added changesets (`@chart-io/core` minor, `@chart-io/react` minor)

---


packages/react/src/lib/components/Plots/Dendrogram/Dendrogram.unit.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Dendrogram } from "./Dendrogram";
+
+describe("Dendrogram", () => {
+    const data = [
+        { region: "North", product: "Widgets", sales: 5 },
+        { region: "North", product: "Gadgets", sales: 5 },
+        { region: "South", product: "Widgets", sales: 10 },
+    ];
+
+    it("should render without needing an external RectangularChart wrapper", async () => {
```
IPWright83: This test name.makes no sense given this component no longer exists

packages/react/src/lib/components/Plots/Dendrogram/Dendrogram.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Chart, IChartProps, IChartRef } from "../../Chart";
+import { LegendOverlay } from "../../LegendOverlay";
+import { TooltipOverlay } from "../../TooltipOverlay";
+import { ZoomBreadcrumb } from "../../ZoomBreadcrumb";
+
+import { DendrogramPlot, IDendrogramPlotProps } from "./DendrogramPlot";
+
+export interface IDendrogramProps
+    extends Omit<IChartProps, "children">,
+        Omit<IDendrogramPlotProps, "useCanvas" | "onMouseOver" | "onMouseOut" | "onClick"> {
```
IPWright83: Why are we omitting useCanvas and the events?

It should be possible to tell a whole chart to prefer canvas like the XYChart (check pies and treemaps behave this way too). I think the mouse events are needed too so people can hook into event behaviour?

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogram.tsx
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Represents a RadialDendrogram chart, the polar equivalent of `<Dendrogram>` - a tree of nodes built
+ * from the fields listed in `categories`, radiating outward from the center with every leaf aligned at
+ * the same radius. A self-contained chart: no need to wrap it in `<RadialChart>` yourself. Set
+ * `zoomable` to let a click on a node zoom in and refocus the layout on its subtree, and `breadcrumb`
+ * to also show the current zoom path as a clickable trail
+ * @param  props       The set of React properties
+ * @return             The RadialDendrogram component
+ */
+export const RadialDendrogram = forwardRef<IChartRef, IRadialDendrogramProps>(
```
IPWright83: Could we remove this and have a radial flat (default false) on th dendogram that renders this inline instead?

packages/react/src/lib/components/RadialChart/RadialChart.tsx
```diff
@@ -16,24 +17,33 @@ export interface IRadialChartProps extends IChartProps {
...
 }
 
 /**
  * Represents a radial chart. This is the polar equivalent of the `<XYChart>`, and is
- * intended to wrap `<Pie>`, `<Donut>` or `<StackedDonut>` plots
+ * intended to wrap `<Pie>`, `<Donut>`, `<StackedDonut>` or `<RadialDendrogram>` plots
  */
-export const RadialChart = forwardRef<IChartRef, IRadialChartProps>(({ children, centerValue, ...props }, ref) => {
-    const showCenterValue = centerValue ?? hasCenterHolePlot(children);
+export const RadialChart = forwardRef<IChartRef, IRadialChartProps>(
```
IPWright83, line 32: Can we ensure the stacked donut support ls.zooming and breadcrumbs too

packages/react/src/lib/components/Plots/Dendrogram/Dendrogram.unit.tsx
IPWright83: Good catch, renamed to "should render without needing an external Chart wrapper" in 706c019.

---

packages/react/src/lib/components/Plots/Dendrogram/Dendrogram.tsx
IPWright83: They weren't actually missing - `useCanvas`/`onMouseOver`/`onMouseOut`/`onClick` are already declared on `IChartProps` (only "children" is omitted from that extends clause), and `Chart` clones them onto every child (including the plot) via `getChildrenWithProps`, exactly like Pie/Treemap. The second `Omit<IDendrogramPlotProps, "useCanvas" | ...>` was only there to avoid a duplicate-member declaration between the two extended interfaces, but it read as if those props were being stripped from the public API. Removed that Omit entirely in 706c019 - both types declare them identically so there's no conflict, and the props table is now accurate.

---

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogram.tsx
IPWright83: Done - `RadialDendrogram` is gone, `<Dendrogram radial={true}>` now renders the same tree radiating outward from the center via `RadialChart`+`RadialDendrogramPlot` internally (which stays as an internal building block for anyone composing their own chart). Stories/docs/tests merged into Dendrogram's accordingly. Pushed in 706c019.

---

packages/react/src/lib/components/RadialChart/RadialChart.tsx
IPWright83: Added `zoomable` to `StackedDonut`, reusing the same `useZoom` infrastructure as Treemap/Dendrogram/CirclePacking - clicking a non-leaf slice zooms in and refocuses the layout on its subtree, re-expanding the remaining rings to fill the available radius. Breadcrumb support came for free via `RadialChart`'s existing `breadcrumb` prop, since it just reads the shared zoom-path state regardless of which plot set it - no changes needed there. Also caught and fixed a real bug while wiring this up: the per-ring radius lookup was keyed by relative depth (1..levels) but nodes carry absolute depth, so it crashed as soon as you zoomed into a subtree. Pushed in 706c019.

---

packages/core/src/canvas/renderText.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    // font-size may come back unitless (SVG treats it as user units/px) - canvas requires a unit
+    const fontSizeValue = Number.parseFloat(selection.style("font-size")) || 12;
+    const fontSize = `${fontSizeValue}px`;
+
+    context.font = `${fontSize} ${fontFamily}`;
+    context.textAlign = anchor === "end" ? "right" : anchor === "middle" ? "center" : "left";
+    context.textBaseline = "alphabetic";
+
+    if (overrideColor) {
```
IPWright83: I don't think we need hit testing on labels

packages/react/src/lib/components/Plots/Dendrogram/DendrogramBase.tsx
```diff
@@ -247,9 +266,19 @@ export function DendrogramBase({
             .attr("cy", (d: IDendrogramElement & { type: "node" }) => py(d.node))
             .attr("r", 0);
 
+        enter
```
IPWright83: Can we split out the labels, links and circles into different plots. Each can render to canvas separately.

IPWright83: Only do position calculations once though

packages/core/src/canvas/renderText.ts
IPWright83: Good call - removed. `renderText` now draws nothing at all when `overrideColor` is set, so a label no longer occupies any pixels on the virtual canvas; the node it labels is still clickable/hoverable as normal. Pushed in f86ed52.

---

packages/react/src/lib/components/Plots/Dendrogram/DendrogramBase.tsx
IPWright83: Done - links, node circles and labels are now three separate, normally-typed joins (`.dendrogram-link`/`-node`/`-label`) instead of one heterogeneous union with `.filter()` dispatch. To keep this safe for Canvas (a second `renderCanvas` call would still wipe the first's drawing via `clearRect`), `renderCanvas`/`canvasRenderLoop`/`progressiveCanvasRenderLoop` now accept an array of update transitions and clear+draw them together in one render loop - so the three joins are still passed into a single `renderCanvas` call, just as three separate transitions rather than one merged one. Pushed in f86ed52.

---

IPWright83: Confirmed - the hierarchy build and layout (`d3.cluster`, `allNodes`/`allLinks`, and the `px`/`py`/`radiusFor`/`colorFor` helpers) all still run exactly once per render, shared by all three joins. Only the join/enter/transition step is split three ways.

---

IPWright83: Follow-up: switched this to match how multi-series `<Scatter>` actually does it - each of links/nodes/labels now gets its own `withCanvas`/`withSVG`-wrapped layer (own `<canvas>` element in Canvas mode) instead of sharing one join. `renderCanvas`/`canvasRenderLoop` no longer need the array-of-joins support from the previous commit since each plot only ever renders one join to its own canvas - reverted that part. The hierarchy/layout computation still runs once, in a new `useDendrogramLayout`/`useRadialDendrogramLayout` hook shared by the three sibling plots. Pushed in e026923.

---

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogramLabelsBase.tsx
```diff
@@ -0,0 +1,100 @@
...
+    keyFor: (node: IRadialDendrogramNode) => string;
+}
+
+/**
+ * Renders a RadialDendrogram's node labels. Used internally by `<RadialDendrogramBase>`, which computes
+ * the shared layout this renders from - see `useRadialDendrogramLayout`
+ * @param  props       The set of React properties
+ * @return             The RadialDendrogramLabelsBase component
+ */
+export function RadialDendrogramLabelsBase({
```
IPWright83: Could this just be called `LabelsPlot`? Is there anything in here Dendogram specific? I think this could be a generic component that other plots (e.g. say a Force directed graph could also use).

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogramLinksBase.tsx
```diff
@@ -0,0 +1,86 @@
...
+    cy: number;
+}
+
+/**
+ * Renders a RadialDendrogram's connecting links. Used internally by `<RadialDendrogramBase>`, which
+ * computes the shared layout this renders from - see `useRadialDendrogramLayout`
+ * @param  props       The set of React properties
+ * @return             The RadialDendrogramLinksBase component
+ */
+export function RadialDendrogramLinksBase({ layer, canvas, renderVirtualCanvas, allLinks, linkGenerator, keyFor, cx, cy }: IRadialDendrogramLinksBaseProps) {
```
IPWright83: Could this just be called `LinksPlot`? Is there anything in here Dendogram specific? I think this could be a generic component that other plots (e.g. say a Force directed graph could also use).

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogramNodesBase.tsx
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ * Renders a RadialDendrogram's node circles, and handles hover/click interaction, tooltips and zoom.
+ * Used internally by `<RadialDendrogramBase>`, which computes the shared layout this renders from - see
+ * `useRadialDendrogramLayout`
+ * @param  props       The set of React properties
+ * @return             The RadialDendrogramNodesBase component
+ */
+export function RadialDendrogramNodesBase({
```
IPWright83: Could this just be called `NodesPlot`? Is there anything in here Dendogram specific? I think this could be a generic component that other plots (e.g. say a Force directed graph could also use).

packages/react/src/lib/components/Plots/Dendrogram/DendrogramBase.tsx
```diff
@@ -0,0 +1,158 @@
...
+                focusedNode={focusedNode}
+                zoomPath={zoomPath}
+                zoomTo={zoomTo}
+                zoomable={zoomable}
+                interactive={interactive}
+                onMouseOver={onMouseOver}
+                onMouseOut={onMouseOut}
+                onClick={onClick}
+            />
+            <DendrogramLabels renderVirtualCanvas={renderVirtualCanvas} labels={labels} allNodes={allNodes} px={px} py={py} radiusFor={radiusFor} keyFor={keyFor} />
```
IPWright83: Could these all live in the Dendogram component? The same could be true for the radial version then we can remove these 2 orchestration plots.

IPWright83 (commented):
Can we share the Nodes/Links/Labels plots between both Radial and non-radial variants? Surely the only thing that varies is the positions of the nodes (and maybe some animations/link styles, but these can be passed in as props?)

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogramNodesBase.tsx
IPWright83: Good point - nothing here actually referenced the hierarchy at all, positions/colors/events were already passed in as accessor functions. Extracted a generic `<NodesPlot>` (renders positioned, interactive circles with plain `onMouseOver`/`onMouseOut`/`onClick(item, element, event)` callbacks - no notion of tooltips or hierarchies) alongside `<LinksPlot>` and `<LabelsPlot>`, all three living directly under `packages/react/src/lib/components/Plots/` next to the other shared plot-level utilities (`renderCanvas`, `useTooltip`, `useZoom`) rather than under `Dendrogram/`. `DendrogramPlot`/`RadialDendrogramPlot` now compose them, wiring tooltip/focus/zoom into plain callbacks themselves. Pushed in 1bd8ed9.

---

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogramLinksBase.tsx
IPWright83: Done - now a generic `<LinksPlot>` in `Plots/LinksPlot.tsx`. The one wrinkle is the Canvas primitive needs different `data-*` attributes for linear vs radial link geometry (endpoints vs angle/radius pairs, plus a center-point transform for radial) - handled via a `dataAttrs`/`staticAttrs` prop so the caller supplies whatever its `renderLink`/`renderLinkRadial` primitive needs, keeping the component itself agnostic of cartesian vs polar. Pushed in 1bd8ed9.

---

packages/react/src/lib/components/Plots/RadialDendrogram/RadialDendrogramLabelsBase.tsx
IPWright83: Done - now a generic `<LabelsPlot>` in `Plots/LabelsPlot.tsx`, taking plain x/y/text accessors (plus an optional `textAnchor` accessor, which is all the radial variant needed beyond the linear one). Pushed in 1bd8ed9.

---

packages/react/src/lib/components/Plots/Dendrogram/DendrogramBase.tsx
IPWright83: Done - `DendrogramBase`/`RadialDendrogramBase` are gone. `DendrogramPlot`/`RadialDendrogramPlot` now directly call `useDendrogramLayout`/`useRadialDendrogramLayout` and render the three generic plots, wiring tooltip/focus/zoom into plain `onMouseOver`/`onClick` callbacks themselves - that orchestration layer had become a pure pass-through once each shape got its own Canvas/SVG layer, so removing it was a straightforward simplification once the generic plots existed (see the two comments above). Every existing SVG/Canvas snapshot matched without a single update, confirming this and the genericization didn't change behavior. Pushed in 1bd8ed9.

---

PR #234, opened 2026-08-09
Add a CirclePacking chart

IPWright83:
## Summary

- Adds `<CirclePacking>`, nesting a circle per node of the shared `buildHierarchy` hierarchy inside its parent's circle, sized proportionally to `value`. Unlike `<Treemap>`, every node at every level is drawn (not just leaves), using `d3.pack()` for layout.
- Supports the shared `zoomable`/`breadcrumb` chart-level zoom feature (same pattern as `<Treemap>`, `<Dendrogram>`, `<RadialDendrogram>`): clicking a non-leaf circle zooms in and refocuses the layout on its subtree; clicking the current top-level circle, or a breadcrumb entry, zooms back out.
- Only leaf nodes are labelled. Siblings never overlap in a packed layout, but a dominant child is packed concentrically with its parent, so labelling every level stacked group and leaf labels on exactly the same point - a real visual bug, confirmed both in the plain 3-level story and after zooming. Group names stay discoverable via the tooltip and the breadcrumb trail.
- Fixes a `d3.pack()` re-rooting bug: unlike `d3.cluster()`/`d3.tree()`, `d3.pack()`'s internal layout reads a node's *actual* parent position to compute its own, so laying out a zoomed-in node directly produced `NaN` positions for its whole subtree. Fixed by temporarily detaching (and restoring) the focused node's `.parent` around the `d3.pack()` call, with a permanent regression test covering it.

## Test plan

- [x] `pnpm --filter @chart-io/react types` / `pnpm --filter @chart-io/core types` - clean
- [x] `pnpm --filter @chart-io/react test` (330 tests) / `pnpm --filter @chart-io/core test` (168 tests) - all passing, including a new regression test for the zoom NaN bug
- [x] Visually verified in a built Storybook via headless Chromium: Basic, 3-level, Canvas, and Zoomable (with breadcrumb) stories - zoom in, zoom out via breadcrumb, and label placement all confirmed correct with no NaN positions and no overlapping labels

---


IPWright83:
Rebased this PR onto current `master` (dropping the duplicate Dendrogram/RadialDendrogram commits now that #233 has merged) and added the requested Storybook navigation reorg on top, in a separate branch/PR: #247.

I didn't push directly to this PR's branch (`claude/circle-packing`), so this PR is unchanged - #247 has the rebased + reorganized result ready for review. Let me know if you'd like #234 closed in favor of #247, or if you'd rather I push the rebase here instead.

---

PR #245, opened 2026-08-14
Fix Canvas Donut/Pie hover highlight, make StackedDonut zoomable by default

IPWright83:
## Summary

- **Fix Canvas hover highlight**: on Canvas-rendered `<Pie>`/`<Donut>`/`<StackedDonut>` charts, hovering a slice updated its opacity on the underlying (detached) DOM node as before, but nothing repainted the visible `<canvas>` bitmap to reflect it - unlike SVG, where the browser repaints style changes on its own. Added a `redrawCanvas()` helper that `useFocused()` now triggers for Canvas plots. Verified with a real browser (Playwright) against the Pie/Donut/StackedDonut Canvas stories, sampling the hovered slice's pixel alpha before/after.
- **Trim `Pie.mdx`**: removed a sentence describing `<Pie>` vs `<Donut>` from the intro bullet list, since it duplicated the `<Pie>` section further down the same page.
- **`StackedDonut` zoomable by default**: `zoomable` now defaults to `true`. Clicking the donut's center hole now zooms back out one level (previously there was no way to zoom back out by clicking the chart itself, in either SVG or Canvas mode - `handleCenterClick` hit-testing was added to the Canvas virtual event handling for this). While zoomed in, the center hole also now shows the focused node's name whenever nothing's hovered, instead of showing nothing.
- **`ZoomBreadcrumb` stories**: added Storybook stories + docs (`Components/ZoomBreadcrumb`) demonstrating the breadcrumb trail at different zoom depths, and the fully-zoomed-out (renders nothing) state.

## Test plan

- [x] `pnpm --filter @chart-io/core build && pnpm --filter @chart-io/react run types` - clean
- [x] `jest` - all 509 tests pass across `@chart-io/core` and `@chart-io/react` (added new tests for Canvas center-hole zoom-out and `CenterValueOverlay`'s zoom fallback/click behavior)
- [x] `eslint` on all touched files - clean
- [x] Manually verified in a real browser via Playwright against the built Storybook: Canvas hover highlight (Pie/Donut/StackedDonut), StackedDonut zoom-in/out via slice + center-hole clicks (SVG and Canvas), and the new ZoomBreadcrumb stories

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Q8kUMeQP4VbQoEYx24g8Es

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8kUMeQP4VbQoEYx24g8Es)_

### DIFF
--- a/packages/core/src/canvas/virtual/addEventHandlers.ts
+++ b/packages/core/src/canvas/virtual/addEventHandlers.ts
@@ -1,4 +1,4 @@
-import { eventActions } from "../../store";
+import { chartActions, eventActions } from "../../store";
 import type { IDispatch } from "../../store";
 import { MOUSE_MOVE_THROTTLE } from "../../constants";
 import type { IDatum } from "../../types";
@@ -6,6 +6,18 @@ import type { IDatum } from "../../types";
 import { throttle } from "lodash";
 
 import type { IColorToData, IColorToDataMap } from "./types";
+
+/**
+ * The geometry needed to detect a click on a zoomed-in radial plot's (e.g. `<StackedDonut zoomable>`)
+ * center hole, used to zoom back out one level. Undefined/no center radius means the current plot
+ * doesn't have a clickable center hole
+ */
+export interface IZoomCenter {
+    cx: number;
+    cy: number;
+    centerRadius?: number;
+    path: string[];
+}
 
 /**
  * Crazy interface for TypeScript gymnastics
@@ -82,17 +94,37 @@ const triggerOnMouseOut = (datum: IDatum, element: Element, e: MouseEvent) => {
  * @param  {HTMLElement} canvas              The virtual canvas DOM element
  * @param  {Object}      getColorMap         A map from a color to a datum
  * @param  {Function}    dispatch            The redux dispatch function
+ * @param  {Function}    getZoomCenter       Returns the geometry of the focused node's clickable
+ *                                           center hole, if the current plot has one
  * @returns {Object}                         The set of handlers for cleaning up { clickHandler, moveHandler }
  */
 export const addEventHandlers = (
     canvas: HTMLCanvasElement,
     getColorMap: () => IColorToDataMap,
     dispatch: IDispatch,
+    getZoomCenter?: () => IZoomCenter | undefined,
 ) => {
     let lastDatum = undefined;
     let lastNode = undefined;
 
     const eventContext = canvas.getContext("2d", { willReadFrequently: true });
+
+    /**
+     * Translates a mouse event's page-relative coordinates into ones relative to the canvas.
+     * getBoundingClientRect() is recomputed on every event, rather than cached once, since the
+     * canvas's position can change after the event handlers are wired up (e.g. the page being
+     * scrolled, or the chart's container resizing/reflowing). It's viewport-relative, so it's offset
+     * by the current scroll position to line up with pageX/pageY, which are document-relative.
+     * @param  e   The MouseEventArgs
+     * @return     The coordinates, relative to the canvas
+     */
+    const getCoords = (e: MouseEvent) => {
+        const rect = canvas.getBoundingClientRect();
+        return {
+            x: e.pageX - rect.left - window.scrollX,
+            y: e.pageY - rect.top - window.scrollY,
+        };
+    };
 
     /**
      * Retrieve a datum from the click event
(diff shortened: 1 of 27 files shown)


packages/core/src/canvas/virtual/addEventHandlers.ts
```diff
@@ -82,42 +94,72 @@ const triggerOnMouseOut = (datum: IDatum, element: Element, e: MouseEvent) => {
...
     };
 
+    /**
+     * A click that doesn't land on a datum's rendered color is otherwise invisible to a Canvas plot,
+     * since it has no real DOM elements of its own to click on. A zoomed-in radial plot's (e.g. a
+     * `<StackedDonut zoomable>`) center hole is one such gap - it's never itself rendered as a slice
+     * - so it's handled here instead, by checking whether the click falls within it
+     * @param  e   The MouseEventArgs
+     */
+    const handleCenterClick = (e: MouseEvent) => {
```
IPWright83: I think it'd be better to render an invisible circle to handle this click event for both SVG and canvas

IPWright83: Good call, done in d5dea1c. `StackedDonutBase` now renders an actual (invisible) `<circle class="pie-center">` over the hole as part of its own D3 render, alongside the arcs, instead of `addEventHandlers()` doing bespoke distance-from-center hit-testing. It's a real DOM element for SVG (native click), and gets included in the same Canvas render/virtual-canvas registration as the arcs for Canvas mode - so `addEventHandlers.ts` is back to its original form, and the `IZoomCenter`/`centerRadius` plumbing I'd added to the store is gone.

One thing that surfaced along the way: naively `.merge()`-ing the arcs' and the circle's selections before transitioning didn't actually combine them - `selection.merge()` only fills gaps between an enter/update pair from the *same* join, not two unrelated ones - so the circle was rendering correctly to SVG but never making it into the Canvas render/virtual-canvas pass. Fixed by concatenating their nodes directly and re-wrapping via `d3.selectAll()` before the single shared transition. Verified with Playwright in a real browser against both the SVG and Canvas StackedDonut stories - zoom in, zoom out via the center hole, in both render modes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8kUMeQP4VbQoEYx24g8Es)_

---

packages/react/src/lib/components/CenterValueOverlay/CenterValueOverlay.tsx
```diff
@@ -6,18 +6,22 @@ import { useSelector } from "react-redux";
 /**
  * Represents a center value overlay, which displays the name/value of the currently hovered
  * datapoint in the center of a `<Donut>` or `<StackedDonut>`'s hole, as an alternative to the
- * traditional floating Tooltip
+ * traditional floating Tooltip. While zoomed in (e.g. a `<StackedDonut zoomable>`) and nothing's
```
IPWright83, line 9: Create some stories for this component

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
```diff
@@ -298,28 +299,70 @@ export function StackedDonutBase({
...
                     zoomTo(ancestry(node));
                 }
+            });
+
+        // The focused node itself is never rendered as a slice - it's the empty center hole. While
+        // zoomed in, render an invisible circle there as a real, interactive shape (rather than
+        // special-cased hit-testing), so clicking it zooms back out one level in both SVG and Canvas
+        // mode, the same way any other element on the chart is clicked
+        const showCenterHole = zoomable && zoomPath.length > 0;
+        const centerJoin = d3
```
IPWright83: This would be cleaner by adding one() to our D3 implementation. Can you do so based on https://github.com/d3/d3-selection/pull/300/changes

packages/react/src/lib/components/Plots/renderCanvas.ts
```diff
@@ -36,3 +38,35 @@ export async function renderCanvas(
...
+
+/**
+ * Repaints a Canvas from the current state of its (detached) layer, e.g. after a hover/focus change has
+ * updated a node's style (such as opacity) directly via D3. Canvas is just a bitmap, so unlike SVG,
+ * mutating a node's style has no visual effect until something explicitly repaints it
+ * @param  canvas     The HTML canvas to redraw (a no-op if this plot isn't rendering to one)
+ * @param  width      The width of the chart
+ * @param  height     The height of the chart
+ * @param  layer      The (detached) layer holding the nodes to redraw
+ */
```
IPWright83, line 50: Do existing charts (e.g. scatter) support highlight on canvas as I'd prefer not to add this function if we can avoid it

packages/react/src/lib/components/CenterValueOverlay/CenterValueOverlay.tsx
IPWright83: Added in e82f86c - `CenterValueOverlay.stories.tsx` + `.mdx` under `Components/CenterValueOverlay`, covering: hovering an item, zoomed in with nothing hovered (shows the focused node's name), zoomed in while hovering a slice (hover takes priority), and the empty state (nothing hovered, not zoomed - renders nothing).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8kUMeQP4VbQoEYx24g8Es)_

---

packages/react/src/lib/components/Plots/Pie/StackedDonut/StackedDonutBase.tsx
IPWright83: Done in e82f86c - added `d3.one(selection, name[, className[, present]])` to `@chart-io/core`'s D3 facade (`packages/core/src/d3/one.ts`), based on the linked PR's `selectAll(...).data([null]).join(...)` idiom. I extended it with a `present` flag (default `true`) so the element can be conditionally absent - the linked PR's version always creates exactly one, but we need the center hole to disappear entirely once fully zoomed out.

`StackedDonutBase`'s center-hole hit target is now just:
```ts
const centerUpdate = d3
    .one<Element, Element, unknown>(d3.select(layer.current), "circle", "pie-center", showCenterHole)
    .attr("cx", cx)
    .attr("cy", cy)
    .attr("r", innerRadiusPx)
    ...
```
replacing the hand-rolled select/data/enter/exit/merge block. Added unit tests for `one()` itself (create/reuse/remove/default-present/no-class) in `one.unit.ts`, and confirmed the StackedDonut zoom tests (SVG + Canvas) still pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8kUMeQP4VbQoEYx24g8Es)_

---

packages/react/src/lib/components/Plots/renderCanvas.ts
IPWright83: Checked - no, it doesn't look like any existing Canvas plot supports this today. I tested `<Scatter useCanvas>` live (hovered a point, sampled the pixel's alpha before/during/after): it stays flat the whole time, no highlight. `Scatter`'s `useFocused` (and `Column`'s/`Bar`'s, which follow the same pattern) only sets `opacity` on the underlying DOM node via `d3.select(focused.element).style(...)`, same as `Donut`'s did before this PR - which works for SVG (the browser repaints the style change) but silently does nothing for Canvas (nothing repaints the bitmap). So this isn't duplicating an existing capability - it's a real, currently-unfixed gap everywhere Canvas + hover-highlight are combined, and `redrawCanvas()` is generic enough that `Scatter`/`Column`/`Bar` could adopt it the same way if you want.

Given that, `redrawCanvas()` seems necessary to actually fix the bug this PR is for (Donut/Pie/StackedDonut not highlighting on hover in Canvas mode) - I don't see a way to fix it without something along these lines. Want me to also wire it into Scatter/Column/Bar in this PR for consistency, or leave that as a follow-up and keep this PR scoped to the reported Donut/Pie bug?

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8kUMeQP4VbQoEYx24g8Es)_

---

PR #247, opened 2026-08-15
Rebase CirclePacking chart onto master, reorganize Storybook navigation

IPWright83:
## Summary

This branch rebases #234 (`Add a CirclePacking chart`) onto the current `master` and adds a Storybook navigation reorganization on top.

- **Rebase of #234**: `claude/circle-packing` was branched before #233 (Dendrogram/RadialDendrogram) was merged, so it carried a full duplicate copy of that history. Rebuilt onto current `master` via `git rebase --onto origin/master 0d894cd`, keeping only the 5 CirclePacking-specific commits (`7c7c63d`..`f22ec1d`) and dropping the now-duplicate Dendrogram commits that #233 already delivered. Reviewed the last 10 merged PRs' review threads (#232, #233, #237, #239, #240, #243, #245) to confirm CirclePacking follows the same established patterns: self-contained chart (no `RectangularChart` wrapper), `zoomable`/`breadcrumb` via `useZoom`, chart-level dimensions pulled from `chartSelectors`, docs/mdx kept in sync - all of which #234's existing commits already do.
- **Storybook reorganization**:
  - Removed the `Architecture/Introduction` page (and the now-empty `Architecture` entry from `storySort`).
  - Moved `RadialCharts/*` and `XYCharts/*` under a new top-level `Charts` group (`Charts/RadialCharts/*`, `Charts/XYCharts/*`).
  - Moved `Dendrogram`, `Treemap` and `CirclePacking` out of the catch-all `Plots` group into `Charts/Hierarchical`, alongside the other chart families (confirmed with the requester since CirclePacking wasn't explicitly named but is the same kind of hierarchy-based chart as Dendrogram/Treemap).

## Test plan

- [x] `pnpm --filter @chart-io/core build` / `pnpm --filter @chart-io/core types` / `pnpm --filter @chart-io/react types` - clean
- [x] `pnpm --filter @chart-io/core test` (176 tests) / `pnpm --filter @chart-io/react test` (351 tests) - all passing
- [x] `pnpm --filter @chart-io/react build-storybook` - builds cleanly; verified the generated story index shows `Charts/Hierarchical/{CirclePacking,Dendrogram,Treemap}`, `Charts/RadialCharts/*`, `Charts/XYCharts/*`, and no `Architecture/*` entry

---
_Generated by [Claude Code](https://claude.ai/code/session_01XLvXiPFNZCTMa9M24qyoSJ)_


packages/react/src/lib/components/Plots/CirclePacking/CirclePacking.unit.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { CirclePacking } from "./CirclePacking";
+
+describe("CirclePacking", () => {
+    const data = [
+        { region: "North", product: "Widgets", sales: 5 },
+        { region: "North", product: "Gadgets", sales: 5 },
+        { region: "South", product: "Widgets", sales: 10 },
+    ];
+
+    it("should render without needing an external RectangularChart wrapper", async () => {
```
IPWright83: Just rename to should render correctly?

packages/react/src/lib/components/Plots/CirclePacking/CirclePackingBase.tsx
```diff
@@ -0,0 +1,336 @@
...
+}
+
+/**
+ * Represents a CirclePacking plot, nesting a circle per node of the hierarchy built from
+ * `categories` inside its parent's circle, each sized proportionally to `value`. Used internally by
+ * `<CirclePacking>` - use that unless you need to compose the plot into a chart of your own
+ * @param  props       The set of React properties
+ * @return             The CirclePackingPlot component
+ */
+export function CirclePackingBase({
```
IPWright83: Can we re-use the NodesPlot and LabelsPlot for the Dendogram without too many changes? Am I correct in thinking the positions/layout are the key difference?

packages/react/src/lib/components/Plots/CirclePacking/CirclePacking.unit.tsx
IPWright83: Done in 091056d.

---

packages/react/src/lib/components/Plots/CirclePacking/CirclePackingBase.tsx
IPWright83: Yes - positions/layout (`d3.pack()` vs `d3.cluster()`) were the only real difference; the node/label rendering, hover/click/zoom wiring, and Canvas/SVG plumbing were already identical in shape to `DendrogramPlot`.

Done in 091056d, following the same structure as `DendrogramPlot`/`useDendrogramLayout`: added a `useCirclePackingLayout` hook (hierarchy build, pack layout, `px`/`py`/`colorFor`/`keyFor`/`ancestry`/`breadcrumb`, zoom state), and `CirclePackingPlot` now renders `<NodesPlot>`/`<LabelsPlot>` directly - each gets its own Canvas/SVG layer via `withCanvas`/`withSVG`, same as Dendrogram's links/nodes/labels. `CirclePackingBase` is gone entirely. Regenerated the Canvas image snapshots (now 2 canvases - nodes then labels - instead of 1) and updated that test to match; all existing behavior (click-to-zoom, tooltips, legend, label radius threshold) is unchanged and covered by the existing tests.

---

packages/react/src/lib/components/Plots/CirclePacking/__image_snapshots__/circle-packing-plot-unit-tsx-circle-packing-plot-using-canvas-should-render-correctly-1-snap.png
IPWright83, line 1: Was this snapshot supposed to change? Looks like this may have broken due to a opacity/color change?

IPWright83: Good catch, that was a real regression, not intentional. `CirclePackingBase` used to style every node circle with `theme.series.opacity` unconditionally; the generic `NodesPlot` I switched to (shared with `DendrogramPlot`, which never needed this since it always renders at full opacity) had no equivalent, so it silently got dropped in the refactor - nodes were rendering fully opaque instead of at the theme's series opacity (0.7 in the test theme).

Fixed in 53e720d: added an optional `opacity` prop to `NodesPlot` and pass `theme.series.opacity` through from `CirclePackingPlot`. Regenerated the affected image snapshot and confirmed the SVG snapshot now shows `opacity: 0.7` on each node again, matching pre-refactor behavior.

---

PR #249, opened 2026-08-15
Fix Canvas hover highlight on Column/Bar plots

IPWright83:
## Summary

Follow-up to #245's fix for Canvas hover highlighting. That PR fixed `Donut`/`Pie`/`StackedDonut`, and left a review comment asking whether the same gap existed elsewhere ([discussion](https://github.com/IPWright83/chart-io/pull/245#discussion_r3789636349)). It does:

- **`Column`/`Bar` (and their `Grouped`/`Stacked` variants)**: their shared `useFocused` hooks set `opacity` directly on the hovered element's underlying (detached) DOM node, same as `Donut` did before #245 - which works for SVG (the browser repaints the style change) but silently does nothing for Canvas (nothing repaints the bitmap). Wired in the `redrawCanvas()` helper from #245 so hovering a Column/Bar segment on `useCanvas` now actually highlights it, and un-highlights it again on mouse-out.
- **`Scatter`**: checked, but not affected - its `useFocused` only adds/removes markers and droplines, it never touches an element's opacity, so there's no equivalent bug to fix there.

## Test plan

- [x] `pnpm --filter @chart-io/core build && pnpm --filter @chart-io/react run types` - clean
- [x] `pnpm test` - all suites pass (added new tests asserting the canvas bitmap's pixel alpha changes on hover and reverts on exit for `Column` and `Bar`; verified the new tests fail without the `redrawCanvas()` call, confirming they catch the regression)
- [x] `eslint` on all touched files - clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01U2Z2PguXVLMYgcqcA5RpHB

---
_Generated by [Claude Code](https://claude.ai/code/session_01U2Z2PguXVLMYgcqcA5RpHB)_

### DIFF
--- a/packages/react/src/lib/components/Plots/Bar/Bar/BarBase.tsx
+++ b/packages/react/src/lib/components/Plots/Bar/Bar/BarBase.tsx
@@ -48,7 +48,7 @@ export function BarBase({
 
     useLegendItem(x, "square", showInLegend, fillColor);
     const onTooltip = useTooltip({ y });
-    const onFocus = useFocused({ yScale, theme, grouped: false });
+    const onFocus = useFocused({ yScale, theme, grouped: false, canvasRedraw: { canvas, width, height, layer } });
 
     useRender(() => {
         const { bandwidth, offset } = getBandwidthAndOffset(yScale, y, data);
--- a/.changeset/column-bar-canvas-hover-highlight.md
+++ b/.changeset/column-bar-canvas-hover-highlight.md
@@ -1,0 +1,5 @@
+---
+"@chart-io/react": patch
+---
+
+Fixed the hovered item not being highlighted on Canvas-rendered `<Column useCanvas>`, `<GroupedColumn useCanvas>`, `<StackedColumn useCanvas>`, `<Bar useCanvas>`, `<GroupedBar useCanvas>` and `<StackedBar useCanvas>` charts. Hovering updated the item's opacity on its underlying (detached) DOM node as before, but nothing repainted the visible `<canvas>` bitmap to reflect it, so the highlight silently never appeared - unlike SVG, where the browser repaints the style change on its own. This is the same fix already applied to `<Donut>`/`<Pie>`/`<StackedDonut>` in #245.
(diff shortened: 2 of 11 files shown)


packages/react/src/lib/components/Plots/Bar/Bar/BarBase.tsx
```diff
@@ -48,7 +48,7 @@ export function BarBase({
 
     useLegendItem(x, "square", showInLegend, fillColor);
     const onTooltip = useTooltip({ y });
-    const onFocus = useFocused({ yScale, theme, grouped: false });
+    const onFocus = useFocused({ yScale, theme, grouped: false, canvasRedraw: { canvas, width, height, layer } });
```
IPWright83: Could we change this to grab the width and height via a selector instead?

IPWright83: Done in 990ec4c - `Column`'s and `Bar`'s `useFocused` hooks now pull `width`/`height` from the store themselves (`chartSelectors.dimensions`), so callers only need to pass `canvasRedraw: { canvas, layer }`. Updated all six call sites (`Column`/`GroupedColumn`/`StackedColumn`/`Bar`/`GroupedBar`/`StackedBar`Base).

---

packages/react/src/lib/components/Plots/Bar/Bar/BarBase.tsx
```diff
@@ -48,7 +48,7 @@ export function BarBase({
 
     useLegendItem(x, "square", showInLegend, fillColor);
     const onTooltip = useTooltip({ y });
-    const onFocus = useFocused({ yScale, theme, grouped: false, canvasRedraw: { canvas, width, height, layer } });
+    const onFocus = useFocused({ yScale, theme, grouped: false, canvasRedraw: { canvas, layer } });
```
IPWright83: Can we just pass canvas and layer through now. Remove the canvasRedraw?

IPWright83: Done in 06f1441 - dropped the `canvasRedraw` wrapper entirely, `canvas`/`layer` are now passed straight through to `useFocused` as top-level props (matching `width`/`height`, which are already pulled from the store internally). Updated all six call sites.

---

PR #250, opened 2026-08-15
Add baseline accessibility pass (ARIA + keyboard) to charts

IPWright83:
## Summary

Targets the accessibility gap flagged in the competitive field report: *"No ARIA roles/labels or keyboard interaction found anywhere in `packages/react/src`... Highcharts and DevExtreme market accessibility as a headline feature specifically because it's an enterprise procurement requirement (WCAG/Section 508). Right now that segment can't consider chart-io at all."*

This is the P1 "baseline accessibility pass" item — ARIA roles/labels on chart containers and series, plus keyboard focus on interactive marks — scoped to the chart types most dashboards actually ship (Bar/Column families, Scatter, Pie/Donut, StackedDonut).

- **`<Chart>`** (the shared root SVG for every chart type — `XYChart`, `RadialChart`, etc.) gains `title` / `description` / `role` props. `title` renders as the SVG's `<title>` and is wired up via `aria-labelledby`; `description` renders as `<desc>` via `aria-describedby`. Without a `title`, the chart falls back to `aria-label="Chart"` rather than being announced as a nameless graphic. The root SVG is now `tabIndex={0}` and defaults to `role="img"` (overridable).
- **`Bar`, `Column`, `GroupedBar`/`GroupedColumn`, `StackedBar`/`StackedColumn`, `Scatter`, `Pie`/`Donut`, `StackedDonut`** marks (in both SVG and Canvas mode) now get `role="img"`, a data-driven `aria-label` (e.g. `"North / Widgets: 5"` on a StackedDonut slice), and `tabindex={0}` when the plot is `interactive`.
  - Tab-focusing a mark mirrors the existing mouse-hover highlight and fires the same `onMouseOver` callback a mouse hover would.
  - Enter/Space on a focused mark mirrors a click and fires `onClick` — including `StackedDonut`'s zoom-in/zoom-out behaviour and its center-hole "zoom out" hit target.
- **`<Legend>`/`<LegendItem>`** now expose `role="list"`/`role="listitem"` semantics, and the decorative series-colour swatch is marked `aria-hidden` so screen readers only announce the series name.

**Known limitation** (called out in the changeset): keyboard focus doesn't reposition the floating Tooltip — the `aria-label` on the focused mark covers the same information for screen reader users, but sighted keyboard-only users won't see the visual tooltip. Continuous-path plots (`Line`, `Area`, `Radar`, `RadialArea`) aren't covered in this pass; they need a different (scrubber-style) keyboard interaction model and are a reasonable follow-up.

## Implementation note

Canvas-mode hit-testing in `@chart-io/core` (`addEventHandlers`) replays a mark's `mouseover`/`mouseout` handlers directly by reading them off the DOM node (`node.__on.find(h => h.type === "mouseover")`) using the real underlying `mousemove` event — it never dispatches an actual `"mouseover"` event. So `focus`/`blur` are wired as their own separate `.on()` handlers rather than combined with `mouseover`/`mouseout` (an earlier version of this PR tried `.on("mouseover focus", ...)` with an `event.type` check, which broke Canvas-mode tooltips since `event.type` is always `"mousemove"` there — caught by the existing Canvas interaction tests).

## Test plan

- [x] `pnpm --filter @chart-io/react test` — 76 suites / 352 tests pass (snapshots updated for the new DOM attributes)
- [x] `pnpm --filter @chart-io/react types` (`tsc --noEmit`) — clean
- [x] `pnpm --filter @chart-io/react lint` — clean
- [x] `vite build` — builds cleanly
- [x] Added keyboard focus / Enter / Space / non-activation-key tests for `Bar` and `Donut` (SVG + Canvas dispatch paths), plus a new `Chart.unit.tsx` covering `title`/`description`/`role`/`aria-label` fallback behaviour
- [ ] Manual screen reader spot-check (NVDA/VoiceOver) — not done in this environment, worth a pass before merge

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGbeezwTd28jjfsrijQHTq)_


.changeset/chart-accessibility-baseline.md
```diff
@@ -0,0 +1,25 @@
...
+
+- `<Chart>` (and therefore every chart) now accepts `title` and `description` props, rendered as the
+  root SVG's `<title>`/`<desc>` and wired up via `aria-labelledby`/`aria-describedby` so the chart has a
+  real accessible name and description for screen readers. Without a `title`, the chart still falls back
+  to `aria-label="Chart"` rather than being announced as an unlabelled graphic. The root SVG is also
+  `tabIndex={0}` and defaults to `role="img"`, both overridable.
+- Individual marks on the plots listed above now get `role="img"`, a data-driven `aria-label`, and
+  `tabindex={0}` (when the plot is `interactive`), so they're discoverable via Tab and their value is
+  announced without relying on the mouse-driven Tooltip.
+- Focusing a mark (Tab) now mirrors the existing hover highlight and fires the same `onMouseOver`
```
IPWright83: Is this a good idea? A chart on a page could make it very difficult to tab past (e.g. if it had hundreds of data points)

IPWright83: Good catch - not a good idea as it stood. Fixed in 3493758: replaced the flat `tabindex={0}` on every mark with a roving tabindex (WAI-ARIA APG pattern, same approach Highcharts' accessibility module uses). Each plot instance now has exactly one Tab stop; Left/Right (or Up/Down) arrow keys move that Tab stop - and the DOM focus - between marks, Home/End jump to first/last. So a chart with hundreds of points is still one Tab press to reach and then arrow-key-navigable, not hundreds of Tab presses.

Implementation is a shared `applyRovingTabIndex()` helper (with its own unit tests) wired into all nine plots. `StackedDonut`'s center-hole "zoom out" hit target keeps a plain `tabindex={0}` since it's a single control, not a series of marks. Updated the changeset to describe the new behaviour.


---

PR #252, opened 2026-08-15
Add a WordCloud chart

IPWright83:
## Summary

- Adds `<WordCloud>`: sizes each word in a flat dataset's `category` field proportionally to `value`, packing words - largest first, working outward from the center along an Archimedean spiral - into the available space without overlapping. Set `rotate` to alternate every other placed word between horizontal and vertical.
- Adds `computeWordCloudLayout` to `@chart-io/core`, following the same pattern as `buildHierarchy`: a hand-rolled, fully synchronous, `<canvas>`-measured layout algorithm (rather than a dependency), so it fits the same synchronous D3 render cycle every other plot uses. Words that can't be placed without overlapping are dropped and logged as a new `W010` warning; negative values reuse the existing `W009` warning (its doc page was broadened to mention `<WordCloud>` alongside the hierarchical charts).
- `<WordCloud>` is a self-contained chart - like `<Treemap>`/`<CirclePacking>`, it accepts `data`/`width`/`height` directly, no wrapper needed. `WordCloudBase` renders via its own D3 join (mirroring `DonutBase`/`TreemapBase`) since per-word font-size, rotation and color are unique enough not to fit the generic `NodesPlot`/`LabelsPlot`.
- `renderText()` in `@chart-io/core`'s Canvas renderer gained rotation support (reads a new `data-rotate` attribute) so Canvas rendering matches the SVG output for rotated words.
- Adds Storybook stories/docs (`Charts/Miscellaneous/WordCloud`), a `word_frequency_dataset` for the stories, and a changeset.

## Test plan

- [x] `computeWordCloudLayout` unit tests (core): font-size scaling, placement order/determinism, non-overlap/bounds invariants, rotation, negative-value (`W009`) and can't-fit (`W010`) warnings, custom `measureText` override
- [x] `WordCloud`/`WordCloudPlot` unit tests (react): SVG + Canvas rendering (incl. image snapshots), interactivity (mouseover/click/interactive=false), rotation, custom `measureText`
- [x] `pnpm test` green across `@chart-io/core` and `@chart-io/react` (79 suites / 363 tests)
- [x] `tsc --noEmit` and `eslint` clean in both packages
- [x] Manually verified the Basic/Rotated/Canvas stories in a running Storybook instance (screenshots), confirming layout, tooltips, and Canvas/SVG parity

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01G8AQTSNP1smsXUGNjcSY89

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8AQTSNP1smsXUGNjcSY89)_


packages/react/src/lib/components/Plots/WordCloud/__image_snapshots__/word-cloud-plot-unit-tsx-word-cloud-plot-using-canvas-should-render-correctly-1-snap.png
IPWright83, line 1: I think this needs more word on to test, but also these words are overlapping? That shouldn't happen right?

packages/react/src/lib/components/Plots/WordCloud/__image_snapshots__/word-cloud-plot-unit-tsx-word-cloud-plot-using-canvas-should-render-correctly-2-snap.png
IPWright83, line 1: I don't think this snapshot is supposed to empty is it?

packages/react/src/lib/components/Plots/WordCloud/WordCloud.tsx
IPWright83, line 1: I think a wordcloud should have an option of containing inside bubbles.

packages/react/src/lib/components/Plots/WordCloud/WordCloud.unit.tsx
```diff
@@ -0,0 +1,36 @@
...
+describe("WordCloud", () => {
+    const data = [
+        { word: "chart", count: 50 },
+        { word: "data", count: 30 },
+        { word: "plot", count: 10 },
+    ];
+
+    const measureText = () => ({ width: 40, height: 14 });
+
+    it("should render without needing an external chart wrapper", async () => {
```
IPWright83: Stop mentioning external chart wrapper

packages/react/src/lib/components/Plots/WordCloud/WordCloud.unit.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import { render } from "@testing-library/react";
+
+import { wait } from "../../../testUtils";
+import { WordCloud } from "./WordCloud";
+
+describe("WordCloud", () => {
```
IPWright83, line 8: Should have tests for Canvas too, and events on both SVG and Canvas.

packages/react/src/lib/components/Plots/WordCloud/__image_snapshots__/word-cloud-plot-unit-tsx-word-cloud-plot-using-canvas-should-render-correctly-1-snap.png
IPWright83: Good catch — real bug, not just a bad screenshot. The layout algorithm itself is fine (it always reserves non-overlapping boxes), but the test's `measureText` mock returned a *fixed* `{ width: 8*text.length, height: 14 }` regardless of `fontSize` — so a 60px-tall "chart" was measured as if it were 14px tall, and the algorithm correctly avoided overlapping that (wrong, tiny) box while the actually-rendered 60px glyphs spilled outside it. Fixed the mock to scale with `fontSize`, and picked sane `minFontSize`/`maxFontSize` per test so words still fit their small test canvases. Snapshots regenerated in e09ef2b — no overlap now.

---

packages/react/src/lib/components/Plots/WordCloud/__image_snapshots__/word-cloud-plot-unit-tsx-word-cloud-plot-using-canvas-should-render-correctly-2-snap.png
IPWright83: You're right, that was a real bug — not supposed to be empty. `renderText()` skipped drawing entirely during the hit-testing pass (`if (overrideColor) return;`), which is correct for a label that sits *on top of* an already-interactive shape (a `CirclePacking` node's circle, a `Treemap` cell) but wrong for `WordCloud`, where the text itself is the only shape — there's no backing rect/circle underneath it. So its virtual canvas came out completely blank, and Canvas-mode hover/click could never find a color to match against. Fixed `renderText()` to paint a solid glyph silhouette in `overrideColor` during the hit-testing pass, same as every other renderer (rect/circle/arc/polygon). Regenerated snapshot in e09ef2b now shows the actual hit-test regions, and I added Canvas mouseover/click tests (previously only SVG events were covered) that pass against the fix.

---

packages/react/src/lib/components/Plots/WordCloud/WordCloud.unit.tsx
IPWright83: Done in e09ef2b — reworded to "should render correctly", matching the other self-contained charts' (Treemap/CirclePacking) test naming.

---

packages/react/src/lib/components/Plots/WordCloud/WordCloud.unit.tsx
IPWright83: Added in e09ef2b: `WordCloud.unit.tsx` now has a Canvas render test plus SVG mouseover/mouseout/click and Canvas mouseover (virtual-canvas hit-test) coverage, on top of the existing SVG/Canvas/event coverage already in `WordCloudPlot.unit.tsx`.

---

packages/react/src/lib/components/Plots/WordCloud/WordCloud.tsx
IPWright83: Interesting idea — I want to make sure I build the right thing before taking it on, since it's a fair bit of new layout work. Could you clarify which of these you mean (or something else)?

1. Each word gets its own individual bubble/circle behind it (a bit like a `CirclePacking` node, but sized to fit the word inside).
2. The whole cloud is masked into an overall bubble/circle shape, instead of filling the rectangular plot area.

Happy to scope and build either as a follow-up once I know which - didn't want to guess and build the wrong one into this PR.

---

IPWright83:
`unit` and `lint` are both green on 3d58f77. The remaining "UI Tests"/"UI Review" failures are Chromatic reporting 1 changed story - expected for a brand-new `WordCloud` story with no existing baseline, and needs a maintainer to accept it in the [Chromatic review](https://www.chromatic.com/review?appId=616ea3059f84e2003aa9a458&number=252&type=linked) (not something a code push resolves).

---

packages/react/src/lib/components/Plots/WordCloud/WordCloud.tsx
IPWright83: The first option, using a CirclePacking node would work nicely, obviously they all have different sizes but I don't know if a CirclePacking layout is always the best.

IPWright83: Got it, thanks. To be clear on scope: I wouldn't reuse the actual `CirclePacking` layout algorithm (that packs circles into each other to *fill* space efficiently, which fights against the spiral word-placement this chart already does) - just a circle drawn to fit around each word's own already-computed position/size, sized off its measured bounding box (`Math.hypot(width, height) / 2` or similar), same color as the word, rendered behind it. So it's a per-word visual embellishment on top of the existing layout, not a different overall placement algorithm - hopefully that addresses the "is CirclePacking always best" concern, since it isn't actually CirclePacking under the hood.

I'll build this as a follow-up PR (an optional `bubbles` prop) rather than growing this one further while it's already mid-review - let me know if you'd rather have it folded in here instead.

---

packages/react/src/lib/components/Plots/WordCloud/__image_snapshots__/word-cloud-plot-unit-tsx-word-cloud-plot-using-canvas-should-render-correctly-2-snap.png
IPWright83: Would it be better to wrap the text somehow in a shape with a hit detection on it, that way gaps in letters won't flicker.

IPWright83: Yes, exactly right, and done in 0fc33ab. `<WordCloud>` now stamps each word's own (unrotated) measured width/height onto its `<text>` node as `data-width`/`data-height`. When `renderText()` sees those during the hit-testing pass, it paints a solid rectangle (sized and rotated to match the word's own already-computed layout box) instead of the glyph outlines - so hovering a gap inside/between letters no longer drops out of the hit color. Falls back to the old glyph-fill behavior when those attributes aren't present, so it doesn't touch `CirclePacking`/`Treemap`/`Dendrogram`'s labels. Updated snapshot shows solid boxes now instead of letter shapes.

---

PR #253, opened 2026-08-15
Add a Chord chart

IPWright83:
## Summary

- Adds `<Chord>`, showing the flow between nodes (built from `source`/`target`) as ribbons connecting arcs arranged around a circle, each arc sized proportionally to its node's total flow. A self-contained chart - like `<CirclePacking>`, `<Treemap>` and `<Dendrogram>` - accepting chart-level props (`data`/`width`/`height`) directly, with Canvas support via `useCanvas`.
- `useChordLayout` builds the node list/matrix from a flat `source`/`target`/`value` dataset, runs `d3.chord()`, and computes group/ribbon geometry, colors and label positions - shared between the shapes and labels layers, the same way `useCirclePackingLayout`/`useRadialDendrogramLayout` work.
- `ChordShapesPlot` is a bespoke renderer (like `Donut`/`StackedDonut`) for the group arcs and ribbons, since neither the generic `NodesPlot` nor `LinksPlot` supports per-shape transition/interpolation or a `data-path-type` for Canvas.
- Adds a `"ribbon"` Canvas path type (`renderRibbon`, mirroring `renderArc`) and the `d3-chord` dependency/barrel export in `@chart-io/core`.
- Promotes `interpolateArc` from `Pie/` to the shared `Plots/` level (now used by `ChordShapesPlot` too) and adds a sibling `interpolateRibbon`.
- Adds `Charts/Relational/Chord` Storybook stories and docs, plus an illustrative trade-flows sample dataset.
- Adds a changeset.

## Test plan

- [x] `pnpm --filter @chart-io/core test` - all passing (175/176, 1 pre-existing skip)
- [x] `pnpm --filter @chart-io/react test` - all passing (364/364), including new `Chord.unit.tsx`/`ChordPlot.unit.tsx` (SVG + Canvas, with image snapshots for the new ribbon rendering)
- [x] `pnpm --filter @chart-io/react run types` / `lint` - clean
- [x] `pnpm --filter @chart-io/react run build-storybook` - builds successfully, including the new Chord stories/docs
- [x] Manually inspected the generated Canvas image snapshot to confirm the rendered chord diagram looks correct (arcs + ribbons, not just a shape count check)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FUGvu1jn9EF7B5op6hjP6F)_


packages/react/src/lib/components/Plots/Chord/Chord.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    padAngle={padAngle}
+                    cornerRadius={cornerRadius}
+                    sort={sort}
+                    colors={colors}
+                    labels={labels}
+                    showInLegend={showInLegend}
+                    interactive={interactive}
+                />
+                <TooltipOverlay onlyNearest={true} />
+                <LegendOverlay />
```
IPWright83, line 54: What does a legend show here?

packages/react/src/lib/components/Plots/Chord/ChordPlot.unit.tsx
```diff
@@ -0,0 +1,178 @@
...
+            });
+
+            await wait();
+
+            const labels = Array.from(container.querySelectorAll("text.chord-label")).map((el) => el.textContent);
+            expect(labels).toEqual(expect.arrayContaining(["A", "B", "C"]));
+        });
+    });
+
+    describe("using Canvas", () => {
```
IPWright83, line 145: Looks like we're missing interaction tests on canvas?

packages/react/src/lib/components/Plots/Chord/Chord.tsx
```diff
@@ -0,0 +1,62 @@
...
+        },
+        ref,
+    ) => {
+        return (
+            <Chart ref={ref} {...chartProps}>
+                <ChordPlot
+                    source={source}
+                    target={target}
+                    value={value}
+                    innerRadius={innerRadius}
```
IPWright83: Can we calculate an inner and outer radius so the user doesn't need to specify them

packages/react/src/lib/components/Plots/Chord/Chord.tsx
IPWright83: One entry per node - its name and the color used for its arc (and any ribbon whose flow originates from it). Since every node's arc is already labelled directly on the diagram (unlike e.g. `<CirclePacking>`, where only some leaves get labelled), the legend was mostly redundant - I've changed `showInLegend` to default to `false` now, same rationale `<Dendrogram>`/`<RadialDendrogram>` use. Still available via `showInLegend={true}` for cases where the on-diagram labels get crowded.

---

packages/react/src/lib/components/Plots/Chord/ChordPlot.unit.tsx
IPWright83: Good catch - added a mouseover test for a group and one for a ribbon on the virtual canvas. Rather than hand-deriving the pixel coordinates, each renders an identical SVG version first, reads the real `data-*` geometry attributes off the target shape, and computes a point inside it the same way the shape itself gets drawn (arc midpoint/mid-radius for a group; just inside the ring at the ribbon's source edge for a ribbon) - so the tests stay correct if the layout math ever changes.

---

packages/react/src/lib/components/Plots/Chord/Chord.tsx
IPWright83: Done - replaced `innerRadius`/`outerRadius` with a single `thickness` prop (default `20`px). The outer radius is now derived automatically from the available plot radius, shrinking to leave room for labels when `labels` is set (same idea as `<Dendrogram radial>`'s dynamic label-space reservation), and the inner radius is just `thickness` inside that - so there's no longer a need to specify or reason about two radii.

---

PR #255, opened 2026-08-16
Add Funnel and Pyramid charts

IPWright83:
## Summary

- Adds `<Funnel>`: stacks one trapezoid segment per row of data (`category`/`value`), each segment's width proportional to its value, narrowing from the widest at the top down to the narrowest at the bottom - a classic sales/conversion funnel.
- Adds `<Pyramid>`: a `<Funnel>` turned upside down (`inverted`), so the widest segment forms the flat base at the bottom instead - mirrors the `<Pie>`/`<Donut>` relationship (a thin wrapper around the shared base with one prop fixed).
- Both are self-contained charts, matching the `<Treemap>`/`<CirclePacking>`/`<Dendrogram>` pattern: they accept chart-level props (`data`, `width`, `height`, `plotMargin`, etc.) directly and internally wrap `<Chart>` + a `FunnelPlot` (SVG/Canvas split via `withSVG`/`withCanvas`) + `<TooltipOverlay>`/`<LegendOverlay>`, so no external wrapper chart is needed.
- `FunnelBase` follows the same D3 data-join/render pattern as `DonutBase`/`TreemapBase`: `category`/`value`/`sort`/`colors`/`interactive`/`showInLegend` props, `ensureValuesAreUnique`, tooltip/focus/legend hooks, and `renderCanvas` for the Canvas path. Segments animate in/out via a `points` attribute tween (new segments collapse to a zero-width line and grow into place).
- Promoted Radar's `interpolatePoints` helper (linear interpolation between two SVG `points` strings) from `Plots/Radar/` up to the shared `Plots/` level, alongside the existing `renderCanvas`/`useFocused`/`useTooltip` helpers, since `Funnel` needed the exact same generic logic for its polygon segments.
- Adds Storybook stories/docs (`Funnel.stories.tsx`/`Funnel.mdx`) covering both components, Canvas rendering, sorting, and data-order preservation, plus unit tests (including image snapshots for the Canvas path) and a changeset.

## Test plan

- [x] `pnpm exec jest` - full `@chart-io/react` suite passes (80 suites / 364 tests, including new Funnel/Pyramid/FunnelPlot specs and image snapshots)
- [x] `pnpm exec tsc --noEmit` - clean
- [x] `pnpm exec eslint` on changed files - clean
- [x] `pnpm exec vite build` - package builds cleanly
- [x] `storybook build` - new `Funnel`/`Pyramid` stories build without error

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dr86sXYEo1Na8jf5oPBqJB)_


packages/react/src/lib/components/Plots/Funnel.mdx
```diff
@@ -0,0 +1,71 @@
...
+<Meta title="Charts/Funnel/Funnel" />
+
+# Funnel Plots
+
+There are two closely related components for stacked, tapering plots, these include:
+
+-   `<Funnel>` - one trapezoid segment per row of data, narrowing from the widest at the top down to the narrowest at the bottom
+-   `<Pyramid>` - a `<Funnel>` turned upside down, so the widest segment forms the flat base at the bottom and the narrowest sits at the top
+
+Both are self-contained charts - unlike `<Bar>`/`<Line>` (which need an `<XYChart>` wrapper) or
```
IPWright83: We don't need this comment

packages/react/src/lib/components/Plots/Funnel.mdx
```diff
@@ -0,0 +1,71 @@
...
+| `padding`       | `number`   | `2`                  | The gap, in pixels, to leave between each segment.                                                             |
+| `sort`          | `boolean`  | `false`               | Should the segments be sorted by value (descending), rather than using the order of the data?                 |
+| `inverted`      | `boolean`  | `false`               | Flips the funnel upside down, so the widest segment forms the flat base instead - use `<Pyramid>` for this.   |
+| `colors`        | `string[]` | Derived from Theme   | Override the colours used for each category.                                                                  |
+| `interactive`   | `boolean`  | `true`                | Whether the plot should be interactive. Setting this to false will disable tooltips from this plot            |
+| `showInLegend`  | `boolean`  | `true`                | Whether the plot should feature in the legend or not.                                                         |
+
+`<Funnel>` also accepts the usual chart-level props - `data`, `width`, `height`, `plotMargin`,
```
IPWright83, line 35: Remove mentions of a self contained chart

packages/react/src/lib/components/Plots/Funnel.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+By default segments keep the order that they appear in the data - useful when your data already
+represents the stages of a process in the right order. Set `sort` to order them by value (descending)
+instead.
+
+<Canvas of={FunnelStories.CustomOrder} />
+
+### Negative values
+
+A segment's width can't represent a negative value. If the dataset contains negative values in the
```
IPWright83, line 54: Do we use the no negative warning check that already exists?

packages/react/src/lib/components/Plots/Funnel/FunnelPlot.unit.tsx
```diff
@@ -0,0 +1,186 @@
...
+            await wait();
+
+            expect(console.warn).toHaveBeenCalled();
+
+            (console.warn as jest.Mock).mockRestore();
+        });
+    });
+
+    describe("using Canvas", () => {
+        it("should render correctly", async () => {
```
IPWright83, line 175: Could we add interaction tests?

packages/react/src/lib/components/Plots/Funnel/Funnel.unit.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Funnel } from "./Funnel";
+
+describe("Funnel", () => {
+    const data = [
+        { stage: "Visitors", count: 1000 },
+        { stage: "Leads", count: 500 },
+        { stage: "Customers", count: 100 },
+    ];
+
+    it("should render without needing an external chart wrapper", async () => {
```
IPWright83: Remove comments about an external chart wrapper

packages/react/src/lib/components/Plots/Funnel/Funnel.unit.tsx
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import { render } from "@testing-library/react";
+
+import { wait } from "../../../testUtils";
+import { Funnel } from "./Funnel";
+
+describe("Funnel", () => {
```
IPWright83, line 7: There are no tests for events or canvas here...

packages/core/src/utils/checks/checks.unit.ts
```diff
@@ -69,4 +70,22 @@ describe("/utils/checks", () => {
...
+        });
+
+        it("should return false and warn if any values are negative", () => {
+            const spy = jest.spyOn(console, "warn").mockImplementation(jest.fn());
+
+            const data = [{ x: 1 }, { x: -5 }, { x: 5 }];
+
+            expect(ensureNoNegativeValues(data, "x", "unit_test")).toBe(false);
+
+            expect(spy.mock.calls[0][0]).toMatchSnapshot();
```
IPWright83: Don't use a snapshot for a simple string, put the string inline

packages/core/src/utils/checks/ensureNoNegativeValues.ts
```diff
@@ -0,0 +1,24 @@
+import type { IData } from "../../types";
```
IPWright83, line 1: I thought we already had this helper function?

IPWright83: Checked - it didn't already exist as a standalone helper. The `W009` negative-value check was previously inline inside `buildHierarchy()` only (used by `Treemap`/`StackedDonut`/`CirclePacking`), not exported on its own. Your earlier comment on the mdx ("Do we use the no negative warning check that already exists?") is exactly why this got extracted here - `buildHierarchy` now delegates to this same function instead of duplicating the check inline. Happy to rename/move it if you had a different existing utility in mind.

---

